PR #5, opened 2021-02-22
adding university encounters currently implementable

ThomasLipp:


eldritch/encounters.py
```diff
@@ -144,11 +144,101 @@ def Society4(char):
...
+  return events.Gain(char, {"clues": 1})
+def Administration3(char):
+  check =  events.Check(char, "will", -1)
+  retainer = events.StatusChange(char, "retainer")
+  return events.PassFail(char, check, retainer, events.Nothing())
+def Administration4(char):
+  check = events.Check(char, "lore", -2)
+  spell = events.Draw(char, "spells", 1)
+  curse = events.Curse(char)
+  return events.PassFail(char, check, spell, curse)
```
fingoldfish, line 161: The way I read this encounter, you may choose whether or not to actually make this check.

eldritch/encounters.py
```diff
@@ -144,11 +144,101 @@ def Society4(char):
...
+def Administration5(char):
+  return events.ForceMovement(char, "Asylum") # TODO: also have an event
```
fingoldfish, line 163: Should be able to just Sequence(ForceMovement, Encounter)

eldritch/encounters.py
```diff
@@ -144,11 +144,101 @@ def Society4(char):
...
+def Administration6(char):
+  # Administration 6 is identical to Administration 1
+  return Administration1(char)
 def Administration7(char):
   check = events.Check(char, "will", -2)
   gain = events.Gain(char, {"dollars": 8})
   arrest = events.Arrested(char)
   return events.PassFail(char, check, gain, arrest)
```
fingoldfish, line 171: While you're here, can you fix this? It should also give the player the choice of whether they want to make this check.

eldritch/encounters.py
```diff
@@ -144,11 +144,101 @@ def Society4(char):
...
+  spell = events.Draw(char, "spells", 2)
+  unique = events.Draw(char, "unique", 1)
+  cond = events.Conditional(char, check, "successes", {0: move, 1: spell, 2: unique})
+  return events.Sequence([check, cond], char)
+def Library3(char):
+  check = events.Check(char, "lore", -2)
+  die = events.DiceRoll(char, 1)
+  gain = events.Gain(char, {"clues": die})
+  loss = events.Loss(char, {"stamina": 2, "sanity": 2})
+  return events.PassFail(char, check, events.Sequence([die, gain]), loss)
```
fingoldfish, line 190: You should add the character to the Sequence here.

eldritch/encounters.py
```diff
@@ -144,11 +144,101 @@ def Society4(char):
...
+  return events.PassFail(char, prereq, events.Sequence([unique, move], char), events.Nothing())
+def Science4(char):
+  stamina = events.Loss(char, {"stamina": 2})
+  ally = events.DrawSpecific(char, "allies", "Arm Wrestler") # TODO: prereq on the ally being in the deck
+  return events.BinaryChoice(char, "Arm Wrestle?", "Yes", "No", events.Sequence([stamina, ally], char), events.Nothing())
+def Science5(char):
+  check = events.Check(char, "luck", 0)
+  die = events.DiceRoll(char, 1)
+  win = events.SplitGain(char, "stamina", "sanity", die)
+  loss = events.Gain(char, {"stamina": 1})
```
fingoldfish, line 226: While this is correct, I had to do a double-take because the stamina gain was named loss. Can you name it something else?

eldritch/encounters.py
ThomasLipp: hmm, surprised that it worked without that.

eldritch/encounters.py
ThomasLipp: I agree, that's what I get for multitasking

eldritch/encounters.py
ThomasLipp: bleh, now I need to figure out how to populate the encounter deck for a draw.

eldritch/encounters.py
ThomasLipp: done

eldritch/encounters.py
ThomasLipp: Actually, it looks like Encounter takes a location not a location name, so there may need to be more discussion about how to do this (or I might be missing something)

eldritch/encounters.py
ThomasLipp: changed to coffee

eldritch/encounters.py
```diff
@@ -144,11 +144,105 @@ def Society4(char):
...
+def Administration3(char):
+  check =  events.Check(char, "will", -1)
+  retainer = events.StatusChange(char, "retainer")
+  return events.PassFail(char, check, retainer, events.Nothing())
+def Administration4(char):
+  check = events.Check(char, "lore", -2)
+  spell = events.Draw(char, "spells", 1)
+  curse = events.Curse(char)
+  assist =  events.PassFail(char, check, spell, curse)
+  return events.BinaryChoice(char,"Help the professor and his students?", "Yes", "No", assist, events.Nothing())
```
fingoldfish, line 162: I have decided on a 100 char line limit. Suggested:
```suggestion
  return events.BinaryChoice(
      char,"Help the professor and his students?", "Yes", "No", assist, events.Nothing())
```

eldritch/encounters.py
```diff
@@ -144,11 +144,105 @@ def Society4(char):
...
+def Administration6(char):
+  # Administration 6 is identical to Administration 1
+  return Administration1(char)
 def Administration7(char):
   check = events.Check(char, "will", -2)
   gain = events.Gain(char, {"dollars": 8})
   arrest = events.Arrested(char)
-  return events.PassFail(char, check, gain, arrest)
+  deceive = events.PassFail(char, check, gain, arrest)
+  return events.BinaryChoice(char,"Carry on Deception?", "Yes", "No", deceive, events.Nothing())
```
fingoldfish, line 175: Nit: need space between char, and "

fingoldfish (approved):

eldritch/encounters.py
fingoldfish: The sequence can work without a character attached to it. However, if the character happens to go insane when resolving some event, then the events on the stack that will be unwound are only the events that are attached to that specific character. In this case, it's hard to see how gaining clue tokens would make the character go insane, but it's good to keep in mind for future changes.

eldritch/encounters.py
fingoldfish: I'll get back to this later then.

PR #14, opened 2021-04-06
Allow specifying drawing a specific card type

petercombs:
For instance, "Search the deck for a Weapon", or "Search the Deck for a
Tome" (though we don't yet have Tome as a type).

I think we should probably keep using python types instead of the
Item.item_type attribute to allow for inheritance, but I'm open to suggestions.

### DIFF
--- a/eldritch/events.py
+++ b/eldritch/events.py
@@ -567,15 +567,19 @@ class Draw(Event):
       self.drawn = []
       deck = getattr(state, self.deck)
       i = 0
+      decksize = len(deck)
       while len(self.drawn) < self.draw_count:
         i += 1
-        if not deck or i > 100:
+        if not deck :
           break
         top = deck.popleft()
         if self.target_type is None or isinstance(top, self.target_type):
           self.drawn.append(top)
         else:
           deck.append(top)
+
+        if i >= decksize:
+          break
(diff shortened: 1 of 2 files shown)


petercombs:
For instance, http://www.arkhamhorrorwiki.com/Inner_Sanctum "You may search the Unique Item deck and take the first magical weapon you find"

eldritch/events.py
```diff
@@ -565,10 +566,16 @@ def resolve(self, state):
     if self.drawn is None:
       self.drawn = []
       deck = getattr(state, self.deck)
-      for _ in range(self.draw_count):
-        if not deck:
+      i = 0
+      while len(self.drawn) < self.draw_count:
+        i += 1
+        if not deck or i > 100:
```
ThomasLipp, line 572: why the i > 100?  Is this just arbitrary?

petercombs: Yep. In the unlikely event that there are no weapons remaining, it should stop at some point.

ThomasLipp: seems like we should be able to get a deck length (found before the start of the loop) and use that instead.  Will make it more robust to adding additional cards to the deck.  There should also be a test for when none of the item type exist.

PR #16, opened 2021-05-03
Feature/french hill

petercombs:


eldritch/encounters.py
```diff
@@ -129,22 +129,137 @@ def Lodge1(char):
   check = events.Check(char, "lore", -1)
   draw = events.Draw(char, "spells", 2)
   return events.PassFail(char, check, draw, events.Nothing())
+def Lodge2(char):
+  check = events.Check(char, "fight", -1)
+  ally = events.DrawSpecific(char, "allies", "Thief") # TODO: prereq on the ally being in the deck
```
fingoldfish, line 134: Merge your other branch and then update.

eldritch/encounters.py
```diff
@@ -129,22 +129,137 @@ def Lodge1(char):
...
+  prereq = events.AttributePrerequisite(char, "dollars", 3, "at least")
+  pay = events.Loss(char, {"dollars": 3})
+  check = events.Check(char, "will", -1)
+  damage = events.Loss(char, {"stamina": 2})
+  membership = events.MembershipChange(char, True)
+  resist = events.PassFail(char, check, events.Nothing(), damage)
+  move = events.ForceMovement(char, "FrenchHill")
+  choice = events.BinaryChoice(char, "Join the Lodge?", "Yes", "No",
+                               events.Sequence([pay, membership], char), resist)
+  return events.Sequence([events.PassFail(char, prereq, choice, resist), move], char)
```
fingoldfish, line 150: The way I read this event, you move to the street regardless of whether you take the stamina damage or not, but you don't move to the street if you join the order.

eldritch/events.py
```diff
@@ -356,8 +359,9 @@ def resolve(self, state):
     state.event_stack.append(self)
     self.stack_cleared = True
 
-    # TODO: lose half the items and clues.
```
fingoldfish, line 359: You're going to get some merge conflicts. Sorry. You might not get them if you revert your changes to Insane.

eldritch/events.py
```diff
@@ -568,22 +572,26 @@ def resolve(self, state):
...
-        if self.target_type is None or isinstance(top, self.target_type):
-          self.drawn.append(top)
-        else:
-          deck.append(top)
-
-        if i >= decksize:
-          break
-      # TODO: is there a scenario when the player can go insane/unconscious before they
-      # successfully pick a card?
+      if self.draw_count == 'all':
```
fingoldfish, line 575: Instead of "all", you can use float("inf"). Avoids needing to special-case "all".

eldritch/encounters.py
```diff
@@ -129,22 +129,137 @@ def Lodge1(char):
...
+  return events.PassFail(char, check, gain, events.Nothing())
+def Sanctum2(char):
+  check = events.Check(char, "luck", -1)
+  spend = events.Loss(char, {"sanity": 1})
+  success = events.Nothing() # TODO: claim a monster on the board as a trophy
+  nothing = events.Nothing()
+  seq = events.Sequence([
+    spend, events.PassFail(char, check, success, nothing)
+  ])
+  choice = events.BinaryChoice(char, "Cast a banishment spell?", "Yes", "No", seq, nothing)
```
fingoldfish, line 187: Hmm. Does this card actually give you a choice in the matter? It says "spend one sanity" instead of "you may spend one sanity" so it's not clear to me.

eldritch/encounters.py
```diff
@@ -129,22 +129,137 @@ def Lodge1(char):
...
+      2: events.Loss(char, {'stamina': 1}),
+      3: events.Gain(char, {'stamina': 3})
+    }
+  )
+  return events.Sequence([check, cond], char)
+def Witch4(char):
+  return events.Loss(char, {"sanity": 1})
+def Witch5(char):
+  #TODO: A gate and monster appear
+  return events.Nothing()
```
fingoldfish, line 248: Now possible using OpenGate.

eldritch/encounters.py
```diff
@@ -129,22 +129,137 @@ def Lodge1(char):
...
+def Witch6(char):
+  die = events.DiceRoll(char, 1)
+  gain = events.GainOrLoss(char, {"clues": die}, {"sanity": die})
+  return events.Sequence([die, gain], char)
+def Witch7(char):
+  check = events.Check(char, "will", -2)
+  spell = events.Draw(char, "spells", 1)
+  n_items_to_lose = int(len(char.possessions)//2)
```
fingoldfish, line 256: Ah, you can't quite do it this way. You see, the character may spend items (e.g. the lucky cigarette case, research materials) when they make the check, which means that by the time they have failed the check, they may not have the same number of items as they did when this event was created. I think an event to lose half your items is on the roadmap, where you calculate what half is inside of resolve() so as to avoid this problem. You can stick a TODO here for now.

fingoldfish (changes requested):
I still need to look through the unit tests, but mostly LGTM.

eldritch/events.py
petercombs: Fixed here, which should at least make merging the Draw changes simpler.

eldritch/events.py
petercombs: Not as bad as it seemed...

eldritch/encounters.py
petercombs: TODO added. I left the code to avoid breaking anything, and it's mostly not tested anyways.

eldritch/test_encounters.py
```diff
@@ -459,6 +510,169 @@ def testLodge5Fail(self):
     self.assertEqual(self.char.clues, 0)
     self.assertEqual(self.char.place.name, "FrenchHill")
 
+  def testLodge6Join(self):
```
fingoldfish, line 513: Feel free to skip tests for Lodge6 if it's exactly the same as Lodge4

eldritch/test_encounters.py
```diff
@@ -459,6 +510,169 @@ def testLodge5Fail(self):
...
+    self.assertEqual(len(self.state.unique), 0)
+
+  def testSanctum1Pass(self):
+    self.state.event_stack.append(encounters.Sanctum1(self.char))
+    self.char.lore_luck_slider = 0
+    self.state.unique.extend([items.Cross(), items.HolyWater(), items.TommyGun()])
+    with mock.patch.object(events.random, "randint", new=mock.MagicMock(return_value=5)):
+      choice = self.resolve_to_choice(CardChoice)
+    self.assertEqual(choice.choices, ["Cross", "Holy Water", "Tommy Gun"])
+    self.assertIn("Holy Water", choice.choices)
```
fingoldfish, line 569: How can this assertion ever fail if the previous assertion passes?

eldritch/test_encounters.py
```diff
@@ -459,6 +510,169 @@ def testLodge5Fail(self):
...
+    with mock.patch.object(events.random, "randint", new=mock.MagicMock(return_value=4)):
+      self.resolve_until_done()
+    self.assertEqual(self.char.sanity, 2)
+
+  def testSanctum3OneSan(self):
+    self.char.sanity = 1
+    self.state.event_stack.append(encounters.Sanctum3(self.char))
+    choice = self.resolve_to_choice(MultipleChoice)
+    self.assertEqual(choice.choices, [0, 1])
+    choice.resolve(self.state, 0)
```
fingoldfish, line 613: After resolving the choice, make some assertions?

eldritch/test_encounters.py
```diff
@@ -503,6 +775,115 @@ def testWitch2Fail(self):
...
+      self.resolve_until_done()
+    self.assertEqual(self.char.stamina, 5)
+
+  def testWitch4(self):
+    self.char.sanity = 3
+    self.state.event_stack.append(encounters.Witch4(self.char))
+    self.resolve_until_done()
+    self.assertEqual(self.char.stamina, 2)
+
+  def testWitch4(self):
```
fingoldfish, line 822: You have accidentally named these two tests the same - I believe only one of them will run.

eldritch/test_encounters.py
petercombs: What if a cosmic ray hits the memory between when the first and second asserts happen?

eldritch/test_encounters.py
petercombs: But what are the odds that clobbering the first test will mask that the test was asking if the character lost stamina instead of sanity? 
image

eldritch/test_encounters.py
petercombs: I kept one just to make sure that my implementation of just calling Lodge4 was going to work.

eldritch/encounters.py
```diff
@@ -129,22 +129,140 @@ def Lodge1(char):
...
+  curse = events.BlessCurse(char, False)
+  return events.PassFail(char, check, events.Nothing(), curse)
+def Lodge4(char):
+  prereq = events.AttributePrerequisite(char, "dollars", 3, "at least")
+  pay = events.Loss(char, {"dollars": 3})
+  check = events.Check(char, "will", -1)
+  damage = events.Loss(char, {"stamina": 2})
+  membership = events.MembershipChange(char, True)
+  move = events.ForceMovement(char, "FrenchHill")
+  resist = events.PassFail(char, check, move, events.Sequence([damage, move]))
```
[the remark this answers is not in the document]
petercombs: I don't think it needs it to run... is there a scenario I'm not thinking of where having the character attached makes a difference?

PR #17, opened 2021-05-05
Check whether card is available and give alternate reward for allies if not

petercombs:


eldritch/test_encounters.py
```diff
@@ -819,6 +831,19 @@ def testScience4Yes(self):
     self.assertEqual(self.char.stamina, 1)
     self.assertEqual(self.char.place.name, "Science")
 
+  def testScience4Reward(self):
+    self.char.dollars = 3
+    self.state.event_stack.append(encounters.Science4(self.char))
+    # Yeah, yeah, these aren't spells
```
fingoldfish, line 837: What is this comment referring to? The rest of the test looks correct.

fingoldfish (approved):

eldritch/test_encounters.py
petercombs: Copy paste error from Science 2.

PR #18, opened 2021-05-08
Purchase items at list or discount - Event code

petercombs:
The implementation of the encounters and tests thereof will presumably depend on whether this architecture is correct, so I'm going to hold off on writing those for now.

### DIFF
--- a/eldritch/events.py
+++ b/eldritch/events.py
@@ -1024,6 +1024,119 @@ def GainAllyOrReward(character, ally: str, reward: Event):
   return PassFail(character, has_ally, gain_ally, reward)
 
 
+class Purchase(Event):
+
+  def __init__(self, character, deck, draw_count, discount_type="fixed", discount=0, keep_count=1, target_type=None,
+               prompt="Buy items?"):
+    assert deck in {"common", "unique", "spells"}
+    assert discount_type in {"fixed", "rate"}
+    self.character = character
+    self.deck = deck
+    self.prompt = prompt
+    self.draw_count = draw_count
+    self.keep_count = keep_count
+    self.discount_type = discount_type
+    self.discount = discount
+    self.drawn = None
+    self.choice = None
+    self.kept = []
+    self.prices = None
+    self.paid = []
+    self.target_type = target_type
+    self.resolved = False
+
+  def resolve(self, state):
+    if self.resolved:
+      # This should never happen??
+      return True
+
+    if self.choice is not None:
+      if not self.choice.is_resolved():  # This should never happen
+        return False
+      if self.choice.choices[self.choice.choice_index] == 'Nothing':
+        self.resolved = True
+        for card in self.drawn:
+          getattr(state, self.deck).append(card)
+        return True
+      kept_card = self.drawn.pop(self.choice.choice_index)
+      cost = self.prices.pop(self.choice.choice_index)
+      self.kept.append(self.choice.choices[self.choice.choice_index])
+      assert cost <= self.character.dollars
+      self.character.dollars -= cost
+      self.character.possessions.append(kept_card)
+      self.keep_count -= 1
+
+    if self.drawn is None:
+      self.drawn = []
+      deck = getattr(state, self.deck)
+      i = 0
+      decksize = len(deck)
+      while len(self.drawn) < self.draw_count:
+        i += 1
+        if not deck:
+          break
+        top = deck.popleft()
+        if self.target_type is None or isinstance(top, self.target_type):
+          self.drawn.append(top)
+        else:
+          deck.append(top)
+
+        if i >= decksize:
+          break
+
+    if self.keep_count > 0:
+      choices = []
+      available = []
+      unavailable = []
+      self.prices = []
+      for card in self.drawn:
+        price = self.discounted_price(card)
+        if price <= self.character.dollars:
+          available.append(card)
+          self.prices.append(price)
+          choices.append(f"{card.name} for ${price}")
+        else:
+          getattr(state, self.deck).append(card)
+          unavailable.append(f"{card.name} for ${price}")
+      self.drawn = available
+      choices.append("Nothing")
+
+      if unavailable:
+        could_not_afford = ' (Could not afford {})'.format(','.join(unavailable))
+      else:
+        could_not_afford = ''
+      if available:
+        self.choice = CardChoice(self.character, self.prompt + could_not_afford, choices)
+        state.event_stack.append(self.choice)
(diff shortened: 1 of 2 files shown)


eldritch/events.py
```diff
@@ -612,6 +612,118 @@ def finish_str(self):
...
+
+  def finish_str(self):
+    if not self.kept:
+      return f"{self.character.name} kept nothing"
+    return f"{self.character.name} bought " + ", ".join(self.kept)
+
+  def discounted_price(self, card):
+    if self.discount_type == "fixed":
+      return card.price - self.discount
+    elif self.discount_type == 'rate':
```
fingoldfish, line 724: What kind of monster mixes double quotes and single quotes like this?

petercombs: The kind of monster who normally uses [black](https://github.com/psf/black) to format his code. I’m no longer used to paying attention to inconsistencies, since I trust that the formatter will fix it for me.

eldritch/events.py
```diff
@@ -612,6 +612,118 @@ def finish_str(self):
     return f"{self.character.name} keeps " + ", ".join(self.kept)
 
 
+class Purchase(Event):
+
+  def __init__(self, character, deck, draw_count, discount_type="fixed", discount=0, keep_count=1, target_type=None,
```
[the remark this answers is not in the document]
petercombs: How's the new version look?

eldritch/events.py
```diff
@@ -593,24 +575,172 @@ def resolve(self, state):
...
+        unavailable.append(f"{card.name} for ${price}")
+    self.drawn = available
+    choices.append("Nothing")
+
+    if unavailable:
+      could_not_afford = " (Could not afford {})".format(",".join(unavailable))
+    else:
+      could_not_afford = ""
+    if available:
+      self.choice = CardChoice(self.character, self.prompt + could_not_afford, choices)
```
fingoldfish, line 712: The choices for CardChoice have to be the card names (this is so the UI knows which cards to show the user). We probably need to add annotations to CardChoice so that they can see the final price.

eldritch/test_events.py
```diff
@@ -752,6 +752,178 @@ def testCombatChoice(self):
...
+    choice.resolve(self.state, "Nothing")
+    self.resolve_until_done()
+
+    self.assertTrue(buy.is_resolved())
+    self.assertEqual(self.char.dollars, 3)
+    self.assertFalse(self.char.possessions, [])
+    self.assertEqual(len(self.state.common), 1)
+    self.assertEqual(self.state.common[0], food)
+
+  def testPurchaseOneAtListPoor(self):
```
fingoldfish, line 795: Based on the suspected bug above, please add a test where the first one or two items are too expensive for the player, but the last item is affordable, and then have the player try to purchase the affordable item.

Then assert on the item name in the player's possession.

fingoldfish (changes requested):

eldritch/events.py
```diff
@@ -593,24 +575,172 @@ def resolve(self, state):
...
+
+    if self.choice is not None:
+      if not self.choice.is_resolved():  # This should never happen
+        return False
+      if self.choice.choices[self.choice.choice_index] == "Nothing":
+        self.resolved = True
+        for card in self.drawn:
+          getattr(state, self.draw.deck).append(card)
+        return True
+      kept_card = self.drawn.pop(self.choice.choice_index)
```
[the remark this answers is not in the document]
petercombs: Lower down, I set `self.drawn = available`, so this was actually okay. Test added to show it.

eldritch/events.py
```diff
@@ -543,36 +543,18 @@ def finish_str(self):
     return self.character.name + " moved to " + self.character.place.name
 
 
-class Draw(Event):
+class DrawItems(Event):
```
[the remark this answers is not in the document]
petercombs: We could call it DrawSCUSA I suppose, though I don't know that the acronym is as well known as SCUM.

eldritch/events.py
petercombs: Reverted, and created an issue so we remember that we need to do it.

eldritch/test_events.py
```diff
@@ -1167,6 +1167,199 @@ def testCastAndGoInsane(self):
...
+    draw = Purchase(self.char, "common", 1)
```
fingoldfish, line 1211: Nit: please name this variable `buy` instead of `draw` both to indicate its intent and to match all the other tests.

eldritch/test_events.py
```diff
@@ -1167,6 +1167,199 @@ def testCastAndGoInsane(self):
...
+    self.resolve_until_done()
+
+    self.assertTrue(buy.is_resolved())
+    self.assertEqual(self.char.dollars, 0)
+    self.assertEqual(self.char.possessions, [gun, food])
+    self.assertFalse(self.state.common)
+
+  def testPurchaseTwoAtFixedDiscount(self):
+    buy = Purchase(self.char, "common", 2, keep_count=2, discount=1)
+    self.char.dollars = 8
```
fingoldfish, line 1273: I might make this 6 dollars to validate that the check to see if the character can buy something is done using the discounted price instead of the regular price.

eldritch/test_events.py
```diff
@@ -1167,6 +1167,199 @@ def testCastAndGoInsane(self):
...
+    self.state.event_stack.append(buy)
+    choice = self.resolve_to_choice(CardChoice)
+    self.assertEqual(choice.choices, ["Food", "Tommy Gun", "Nothing"])
+    choice.resolve(self.state, "Tommy Gun")
+    self.resolve_until_done()
+
+    self.assertTrue(buy.is_resolved())
+    self.assertEqual(self.char.dollars, 1)
+    self.assertEqual(self.char.possessions, [gun])
+    self.assertEqual(len(self.state.common), 2)
```
fingoldfish, line 1357: If you want, you can also use something like
`self.assertSequenceEqual(...)`

fingoldfish (approved):

eldritch/events.py
```diff
@@ -971,12 +999,119 @@ def is_resolved(self):
...
+    self.prompt = prompt
+    self.keep_count = keep_count
+    self.discount_type = discount_type
+    self.discount = discount
+    self.draw = draw
+    self.drawn = None
+    self.choice = None
+    self.kept = []
+    self.prices = None
+    self.target_type = target_type
```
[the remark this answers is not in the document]
petercombs: I agree.

PR #19, opened 2021-05-09
Sell items at list or discount

petercombs:


eldritch/events.py
```diff
@@ -612,6 +612,87 @@ def finish_str(self):
     return f"{self.character.name} keeps " + ", ".join(self.kept)
 
 
+class Sell(Event):
+
+  def __init__(self, character, target_decks, sell_count=1, discount_type="fixed", discount=0,
```
fingoldfish, line 617: Same idea - let's make this an ItemChoice followed by a Sell. The ItemChoice has a different user interface from MultipleChoice, though I suppose we would have to extend it to include choosing nothing.

eldritch/events.py
```diff
@@ -612,6 +612,87 @@ def finish_str(self):
...
+        if item.deck in self.target_decks:
+          self.sellable.append(item)
+
+    if self.sell_count > 0:
+      choices = []
+      self.prices = []
+      for card in self.sellable:
+        price = self.discounted_price(card)
+        self.prices.append(price)
+        choices.append(f"{card.name} for ${price}")
```
fingoldfish, line 666: In order for CardChoice to display properly in the UI, the choice names have to be the actual card names. But that shouldn't be a problem if you switch to ItemChoice (see above).

eldritch/events.py
```diff
@@ -612,6 +612,87 @@ def finish_str(self):
...
+  def finish_str(self):
+    if not self.sold:
+      return f"{self.character.name} sold nothing"
+    return f"{self.character.name} sold " + ", ".join(self.sold)
+
+  def discounted_price(self, card):
+    if self.discount_type == "fixed":
+      return card.price - self.discount
+    elif self.discount_type == 'rate':
+      return card.price - int(self.discount * card.price) # Discounts round up
```
fingoldfish, line 694: The only time I can see items being sold for half their list price (rounded in any direction) is a mythos card that says to round down. The word "discount" seems weird here since you're selling it - maybe markdown? Maybe discount is fine.

fingoldfish (changes requested):

fingoldfish (changes requested):

eldritch/events.py
```diff
@@ -2006,22 +2058,41 @@ def BinaryChoice(
...
+    if decks is None:
+      decks = {'spells', 'common', 'unique', 'skills', 'allies'}
+    assert decks.issubset({'spells', 'common', 'unique', 'skills', 'allies'}), f"{decks} contains invalid deck"
+    # can one choose a skill in any circumstances?
+    self.decks = decks
+    self.item_type = item_type
 
   def resolve(self, state, choice=None):
+    if self.choices is None:
+      self.compute_choices(state)
```
fingoldfish, line 2075: You should never have to call `compute_choices` - it is handled by the event loop.

eldritch/events.py
```diff
@@ -2006,22 +2058,41 @@ def BinaryChoice(
 
 class ItemChoice(ChoiceEvent):
 
-  def __init__(self, character, prompt):
+  def __init__(self, character, prompt, decks=None, item_type=items.Item):
```
fingoldfish, line 2061: Why do you use both decks and item_type here? I think if you know the deck, then you automatically know the item type, yes? The exception might be weapons and tomes, but I would prefer those be done with getattr(pos, "item_type").
That being said, you don't seem to be using it anywhere, so maybe you can remove it?

eldritch/events.py
```diff
@@ -2006,22 +2058,41 @@ def BinaryChoice(
...
     self.character = character
     self._prompt = prompt
     self.choices = None
+    self.chosen = None
+    if decks is None:
+      decks = {'spells', 'common', 'unique', 'skills', 'allies'}
+    assert decks.issubset({'spells', 'common', 'unique', 'skills', 'allies'}), f"{decks} contains invalid deck"
+    # can one choose a skill in any circumstances?
```
fingoldfish, line 2069: I'm sure there is. At a minimum, one can choose an ally, for those nasty rumors that ask you to discard an ally.

eldritch/events.py
```diff
@@ -2006,22 +2058,41 @@ def BinaryChoice(
...
+      self.compute_choices(state)
     if self.is_resolved():
       return True
     assert all([0 <= idx < len(self.character.possessions) for idx in choice])
     self.resolve_internal(choice)
 
   def resolve_internal(self, choices):
-    self.choices = [self.character.possessions[idx] for idx in choices]
+    assert all(idx in self.choices for idx in choices)
+    self.chosen = list([
```
fingoldfish, line 2083:
```suggestion
    self.chosen = [self.character.possessions[idx] for idx in choices]
```

eldritch/events.py
```diff
@@ -2051,16 +2122,19 @@ def resolve_internal(self, choices):
...
+    self.min_count = count if min_count is None else min_count
 
   def resolve_internal(self, choices):
-    assert self.count == len(choices)
+    assert self.min_count <= len(choices) <= self.count
     super(ItemCountChoice, self).resolve_internal(choices)
 
 
 class SinglePhysicalWeaponChoice(ItemChoice):
+  # def __init__(self, char, prompt):
```
fingoldfish, line 2136: You could make this inherit from ItemCountChoice and override the constructor to supply count and min_count to the ItemCountChoice constructor. Then you could get rid of resolve_internal

eldritch/events.py
```diff
@@ -2070,6 +2144,15 @@ def resolve_internal(self, choices):
       assert chosen.active_bonuses["physical"] or chosen.passive_bonuses["physical"]
     super(SinglePhysicalWeaponChoice, self).resolve_internal(choices)
 
+  def compute_choices(self, state):
```
fingoldfish, line 2147: nice

eldritch/events.py
```diff
@@ -1037,6 +1038,57 @@ def GainAllyOrReward(character, ally: str, reward: Event):
...
+    self.resolved = False
+    self.sold = None
+
+  def resolve(self, state):
+    if self.resolved:
+      # This should never happen??
+      return True
+    for card in self.choice.chosen:
+      price = self.discounted_price(card)
+      self.prices.append(price)
```
fingoldfish, line 1060: Small nitpick - I recommend
```
self.prices = [self.discounted_price(card) for card in self.choice.chosen]
```
before the for loop. Or alternatively, calling self.prices.clear() before entering the loop. If at some point in the future this event were modified so that resolve might be called multiple times, it would be good to have the calculation of self.prices not re-extend the list.

fingoldfish:
If pylint continues to be problematic, feel free to stick
```
# pylint: disable=assignment-from-none
```
on the problematic line. I actually can't reproduce that locally.

fingoldfish (changes requested):
Just a few more nits.

eldritch/events.py
petercombs: There were some combat tests that were breaking because the event was not getting resolved in the usual way. I've fixed the test and removed this workaround.

eldritch/events.py
petercombs: Weapons and Tomes were what I was thinking of, not sure whether there actually is an encounter that would rely on that feature.  Updated the code at L2085 to use getattr.

fingoldfish (approved):

PR #27, opened 2021-05-22
Losing movement on combat or failed evade

petercombs:
Implement by character having an avoid_monsters attribute that gets reset on Upkeep. This can also be used to implement not fighting monsters upon emergence from a gate.

### DIFF
--- a/eldritch/eldritch.py
+++ b/eldritch/eldritch.py
@@ -359,9 +359,9 @@ class GameState:
     interrupts = []
     if isinstance(event, events.MoveOne):
       nearby_monsters = [
-        mon
-        for mon in self.monsters
-        if mon.place == event.character.place and mon not in event.character.avoid_monsters
+          mon
+          for mon in self.monsters
+          if mon.place == event.character.place and mon not in event.character.avoid_monsters
       ]
       if nearby_monsters:
         interrupts.append(events.EvadeOrFightAll(event.character, nearby_monsters))
--- a/eldritch/eldritch.py
+++ b/eldritch/eldritch.py
@@ -360,7 +360,11 @@ class GameState(object):
   def get_interrupts(self, event):
     interrupts = []
     if isinstance(event, events.MoveOne):
-      nearby_monsters = [mon for mon in self.monsters if mon.place == event.character.place]
+      nearby_monsters = [
+        mon
+        for mon in self.monsters
+        if mon.place == event.character.place and mon not in event.character.avoid_monsters
+      ]
       if nearby_monsters:
         interrupts.append(events.EvadeOrFightAll(event.character, nearby_monsters))
     interrupts += sum([char.get_interrupts(event, self) for char in self.characters], [])
--- a/eldritch/eldritch.py
+++ b/eldritch/eldritch.py
@@ -358,11 +358,7 @@ class GameState:
   def get_interrupts(self, event):
     interrupts = []
     if isinstance(event, events.MoveOne):
-      nearby_monsters = [
-          mon
-          for mon in self.monsters
-          if mon.place == event.character.place and mon not in event.character.avoid_monsters
-      ]
+      nearby_monsters = [mon for mon in self.monsters if mon.place == event.character.place]
       if nearby_monsters:
         interrupts.append(events.EvadeOrFightAll(event.character, nearby_monsters))
     interrupts += sum([char.get_interrupts(event, self) for char in self.characters], [])
--- a/eldritch/events.py
+++ b/eldritch/events.py
@@ -230,6 +230,7 @@ class Movement(Turn):
     self.done = False
 
   def resolve(self, state):
+    self.character.avoid_monsters = []
     if self.check_lose_turn():
       return True
     if self.character.delayed_until is not None:
@@ -495,6 +496,7 @@ class MoveOne(Event):
       self.character.place = state.places[self.dest]
       self.character.movement_points -= 1
       self.character.explored = False
+      self.character.avoid_monsters = []
     self.done = True
     self.moved = True
     return True
@@ -2270,8 +2272,12 @@ class EvadeOrFightAll(Sequence):
   def __init__(self, character, monsters):
    self.monsters = monsters
    self.character = character
-   super(EvadeOrFightAll, self).__init__([
-     EvadeOrCombat(character, monster) for monster in monsters], character)
+   super(EvadeOrFightAll, self).__init__(
+       [
+         EvadeOrCombat(character, monster)
+         for monster in monsters
+         if monster not in character.avoid_monsters
+       ], character)
 
   def start_str(self):
     return f"{self.character.name} must evade or fight all of: " \
@@ -2436,7 +2442,7 @@ class EvadeRound(Event):
       return False
     if self.check.successes >= 1:
       self.evaded = True
-      self.character.avoid_monsters = True
+      self.character.avoid_monsters.append(self.monster)
       return True
     self.damage = Loss(
         self.character, {"stamina": self.monster.damage("combat", state, self.character)})
--- a/eldritch/events.py
+++ b/eldritch/events.py
@@ -510,8 +510,7 @@ class MoveOne(Event):
   def finish_str(self):
     if self.moved:
       return f"{self.character.name} moved to {self.dest}"
-    else:
-      return f"{self.character.name} stayed in {self.character.place}"
+    return f"{self.character.name} stayed in {self.character.place}"
 
 
 class GainOrLoss(Event):
@@ -2275,7 +2274,7 @@ class EvadeOrFightAll(Sequence):
   def __init__(self, character, monsters):
     self.monsters = monsters
     self.character = character
-    super(EvadeOrFightAll, self).__init__(
+    super().__init__(
         [
             EvadeOrCombat(character, monster)
             for monster in monsters
@@ -2460,7 +2459,8 @@ class EvadeRound(Event):
   def finish_str(self):
     if self.evaded:
       return f"{self.character.name} evaded a {self.monster.name}"
-    return f"{self.character.name} did not evade the {self.monster.name} and lost any remaining movement"
+    return (f"{self.character.name} did not evade the {self.monster.name}"
+            + "and lost any remaining movement")
 
 
 class CombatRound(Event):
--- a/eldritch/events.py
+++ b/eldritch/events.py
@@ -297,6 +297,10 @@ class CityMovement(ChoiceEvent):
   def prompt(self):
     return "Move somewhere"
 
+  def __resolve__(self, state):
+      self.done = True
+      self.character.avoid_monsters = False
+
   @property
   def choices(self):
     return sorted(self.routes.keys())
@@ -479,10 +483,12 @@ class MoveOne(Event):
     self.character = character
     self.dest = dest
     self.done = False
+    self.moved = None
 
   def resolve(self, state):
     if self.character.movement_points <= 0:
       self.done = True
+      self.moved = False
       return True
     assert self.dest in [conn.name for conn in self.character.place.connections]
     if not (self.character.place.closed or state.places[self.dest].closed):
@@ -490,16 +496,20 @@ class MoveOne(Event):
       self.character.movement_points -= 1
       self.character.explored = False
     self.done = True
+    self.moved = True
     return True
 
   def is_resolved(self):
     return self.done
 
   def start_str(self):
-    return ""
+    return f"{self.character.name} moving from {self.character.place.name}"
 
   def finish_str(self):
-    return f"{self.character.name} moved to {self.dest}"
+    if self.moved:
+      return f"{self.character.name} moved to {self.dest.name}"
+    else:
+      return f"{self.character.name} did not have enough movement points, and stayed in {self.character.place.name}"
 
 
 class GainOrLoss(Event):
@@ -2258,8 +2268,14 @@ class GateChoice(MapChoice):
 class EvadeOrFightAll(Sequence):
   
   def __init__(self, character, monsters):
+   self.monsters = monsters
+   self.character = character
    super(EvadeOrFightAll, self).__init__([
      EvadeOrCombat(character, monster) for monster in monsters], character)
+
+  def start_str(self):
+    return f"{self.character.name} must evade or fight all of: " \
+           + ", ".join(mon.name for mon in self.monsters)
 
 
 """
@@ -2383,6 +2399,8 @@ class Combat(Event):
     if self.evade.evaded:
       return True
     if self.combat.defeated:
+      self.monster.place = self.character
+      self.character.possessions.append(self.monster)
       return True
     self.choice = None
     return False
@@ -2419,9 +2437,11 @@ class EvadeRound(Event):
       return False
     if self.check.successes >= 1:
       self.evaded = True
+      self.character.avoid_monsters = True
       return True
     self.damage = Loss(
         self.character, {"stamina": self.monster.damage("combat", state, self.character)})
+    self.character.movement_points = 0
     state.event_stack.append(self.damage)
     return False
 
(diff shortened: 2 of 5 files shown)


eldritch/monsters.py
```diff
@@ -7,7 +9,7 @@ def json_repr(self):
     return self.name
 
 
-class Monster(object):
+class Monster(Asset):
```
petercombs, line 12: I don't remember why I did this... I think it had something to do with accessing the monster's name. Regardless, it seems more consistent with the rest of the codebase.

eldritch/events.py
```diff
@@ -1511,6 +1533,8 @@ def resolve(self, state):
       return False
     if self.check.successes >= self.monster.toughness(state):
       # TODO: take the monster as a trophy
+      # Stand-in until we implement trophy code to allow MoveMultipleThroughMonster test to work
+      self.monster.place = None
```
fingoldfish, line 1537: You set up some trophy code above, inside of the Combat event?

eldritch/test_events.py
```diff
@@ -107,6 +107,34 @@ def testMoveOneSpace(self):
...
+
+    third_choice = self.resolve_to_choice(CombatChoice)
+    self.assertIsNone(third_choice.choices)
+    third_choice.resolve(self.state, [])
+    with mock.patch.object(events.random, "randint", new=mock.MagicMock(return_value=5)):
+      self.resolve_until_done()
+
+    self.assertTrue(movement.is_resolved())
+    self.assertEqual(self.char.place.name, "Easttown")
+    # self.assertEqual(self.char.movement_points, 0)
```
fingoldfish, line 134: Why are you not asserting on the movement points here? I know monster trophies may not be complete, but you set movement points to 0 correctly, right?

eldritch/test_events.py
```diff
@@ -120,6 +148,123 @@ def testMoveMultipleSpaces(self):
...
+
+    self.state.event_stack.append(movement)
+    self.resolve_until_done()
+
+    self.assertTrue(movement.is_resolved())
+    self.assertEqual(self.char.place.name, "Easttown")
+    self.assertEqual(self.char.movement_points, 0)
+
+  def testMoveMultipleThroughMonster(self):
+    self.char.fight_will_slider = 2
```
fingoldfish, line 166: Why did you change the fight_will_slider if you're going to have them pass the combat check anyway?

eldritch/test_events.py
```diff
@@ -120,6 +148,123 @@ def testMoveMultipleSpaces(self):
...
+    self.assertTrue(movement.is_resolved())
+    self.assertEqual(self.char.place.name, "Easttown")
+    # self.assertEqual(self.char.movement_points, 0)
+    # TODO: ^ should this pass?
+    # self.assertIn(cultist, self.char.possessions)
+    # TODO: take the monster as a trophy
+
+  def testMoveMultipleThroughMonsterFailedEvade(self):
+    self.char.fight_will_slider = 2
+    self.char.speed_sneak_slider = 0
```
fingoldfish, line 198: You're setting the speed_sneak_slider to 0, but then asserting that the character has 4 movement points. A bit weird, though I suppose it works if you need the extra sneak to make the test work.

eldritch/test_events.py
```diff
@@ -120,6 +148,123 @@ def testMoveMultipleSpaces(self):
...
+    third_choice = self.resolve_to_choice(CombatChoice)
+    self.assertIsNone(third_choice.choices)
+    third_choice.resolve(self.state, [])
+    with mock.patch.object(events.random, "randint", new=mock.MagicMock(return_value=5)):
+      self.resolve_until_done()
+
+    self.assertTrue(movement.is_resolved())
+    self.assertEqual(self.char.place.name, "Easttown")
+    # self.assertEqual(self.char.movement_points, 0)
+    # TODO: ^ should this pass?
```
fingoldfish, line 192: Okay, I figured this out. When you run the unit tests, it actually starts with turn_phase == "upkeep". That means that when you use EndMovement (a subclass of EndTurn), it actually advances from upkeep to movement. When it advances from upkeep to movement, it resets the character's movement points.

I recommend starting these tests in the movement phase.

fingoldfish (changes requested):

eldritch/events.py
```diff
@@ -121,6 +121,10 @@ class EndUpkeep(EndTurn):
   def __init__(self, character):
     super(EndUpkeep, self).__init__(character, "upkeep")
 
+  def __resolve__(self, state):
```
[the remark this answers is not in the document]
petercombs: Yep.

eldritch/events.py
petercombs: Yes, though that code is wrong too. Probably TakeAsTrophy (or similar) should be an event, so we can hijack it for things like Loot Piñatas, personal stories, etc. As a wise man once said, "Passing information between events is most of what happens in this game."

Probably there should only be one bit where I move the monster, though. Any preferences on it appearing in Combat or CombatRound?

eldritch/events.py
```diff
@@ -168,26 +172,33 @@ def __init__(self, character, dest):
...
   def start_str(self):
-    return ""
+    return f"{self.character.name} moving from {self.character.place.name}"
 
   def finish_str(self):
-    return f"{self.character.name} moved to {self.dest.name}"
+    if self.moved:
+      return f"{self.character.name} moved to {self.dest.name}"
+    else:
+      return f"{self.character.name} did not have enough movement points, and stayed in {self.character.place.name}"
```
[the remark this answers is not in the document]
petercombs: I'm shortening it in case the location is closed, which would also block movement.

eldritch/test_events.py
petercombs: Nah, that was mostly me trying to make changes to it, then finding that it did nothing because of your observation below that we weren't actually in the movement phase.

eldritch/events.py
fingoldfish: I think either Combat or CombatRound works. I can't think of a reason why one would work better than the other.

eldritch/monsters.py
```diff
@@ -28,7 +30,8 @@ def __init__(self, name, movement, dimension, difficulties, damages, toughness,
     assert not damages.keys() - self.DAMAGES
     assert not damages.keys() - difficulties.keys()
     assert not set(attributes) - self.ATTRIBUTES
-    self.name = name
+    # self.name = name
```
fingoldfish, line 33: Comment can be removed.

eldritch/test_events.py
```diff
@@ -111,6 +111,36 @@ def testMoveOneSpace(self):
     self.assertEqual(self.char.place.name, "Easttown")
     self.assertEqual(self.char.movement_points, 3)
 
+  def testMoveOneSpaceToMonster(self):
+    while self.state.turn_phase != "movement":
+      self.state.next_turn()
+    maniac = next(monster for monster in self.state.monsters if monster.name == "Maniac")
+    maniac.place = self.state.places["Easttown"]
+    movement = Movement(self.char, [self.state.places["Easttown"]])
```
fingoldfish, line 119: You'll have to rewrite this to use CityMovement - sorry.

eldritch/events.py
```diff
@@ -1471,8 +1490,10 @@ def resolve(self, state):
       return False
     if self.check.successes >= 1:
       self.evaded = True
+      self.character.avoid_monsters = True
```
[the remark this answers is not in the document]
fingoldfish: Upon further review, I definitely think this belongs in EvadeOrFightAll - there are some monsters where you can lose combat against them instead of successfully evading them.

eldritch/test_events.py
petercombs: I like CityMovement better now that we have advance_to_phase!

petercombs:
Not sure I understood the comment above about "Upon further review, I definitely think this belongs in EvadeOrFightAll - there are some monsters where you can lose combat against them instead of successfully evading them." Does the new method of adding monsters to an ignore_monsters list work better?

eldritch/monsters.py
```diff
@@ -50,6 +53,9 @@ def __init__(
       self._attributes.add("nightmarish")
     self.place = None
 
+  def __repr__(self):
+    return f"<Monster: {self.name} {id(self)} at {self.place}>"
```
fingoldfish, line 57: Probably self.place.name if self.place else nowhere

eldritch/eldritch.py
```diff
@@ -358,7 +358,11 @@ def clear_usables(self):
   def get_interrupts(self, event):
     interrupts = []
     if isinstance(event, events.MoveOne):
-      nearby_monsters = [mon for mon in self.monsters if mon.place == event.character.place]
+      nearby_monsters = [
+          mon
+          for mon in self.monsters
+          if mon.place == event.character.place and mon not in event.character.avoid_monsters
```
fingoldfish, line 364: I notice you are doing the filtering for character.avoid_monsters both here before you create the EvadeOrFightAll event, and inside the `__init__` of EvadeOrFightAll. You probably only need one?

eldritch/events.py
```diff
@@ -2423,9 +2442,11 @@ def resolve(self, state):
       return False
     if self.check.successes >= 1:
       self.evaded = True
+      self.character.avoid_monsters.append(self.monster)
       return True
     self.damage = Loss(
         self.character, {"stamina": self.monster.damage("combat", state, self.character)})
+    self.character.movement_points = 0
```
fingoldfish, line 2449: Nitpick: I would either put this line above or below the use of self.damage so that you don't split the definition from the place where it's appended to the stack.

eldritch/events.py
```diff
@@ -2438,7 +2459,8 @@ def start_str(self):
   def finish_str(self):
     if self.evaded:
       return f"{self.character.name} evaded a {self.monster.name}"
-    return f"{self.character.name} did not evade the {self.monster.name}"
+    return (f"{self.character.name} did not evade the {self.monster.name}"
+            + "and lost any remaining movement")
```
fingoldfish, line 2463: Nit: You will want to add a space between the monster name and the word `and`.

eldritch/events.py
fingoldfish: So I realize I said last time that you could do this in either Combat or CombatRound. And I also realize this is a placeholder. But at a minimum, the code for taking the trophy (adding it the player's possessions) and the code for setting the monster's place to None should live in the same place - otherwise someone is going to be very confused in the future.

eldritch/test_events.py
```diff
@@ -249,6 +250,39 @@ def testMoveOneSpace(self):
...
+    next_choice = self.resolve_to_choice(MultipleChoice)
+    self.assertEqual(next_choice.choices, ["Flee", "Fight"])
+    next_choice.resolve(self.state, "Fight")
+
+    third_choice = self.resolve_to_choice(CombatChoice)
+    self.assertFalse(third_choice.choices)
+    third_choice.resolve(self.state, [])
+    with mock.patch.object(events.random, "randint", new=mock.MagicMock(return_value=5)):
+      movement = self.resolve_to_choice(CityMovement)
+      self.assertFalse(movement.choices)
```
fingoldfish, line 277: Nit: the assertFalse and movement resolution can go outside of the patch block.

eldritch/test_events.py
```diff
@@ -262,6 +296,151 @@ def testMoveMultipleSpaces(self):
...
+    )
+    self.assertFalse(movement.is_resolved())
+    self.assertEqual(self.char.movement_points, 1)
+
+    self.state.event_stack.append(movement)
+    self.resolve_until_done()
+
+    self.assertTrue(movement.is_resolved())
+    self.assertEqual(self.char.place.name, "Easttown")
+    self.assertEqual(self.char.movement_points, 0)
```
fingoldfish, line 313: Since you added the `moved` attribute in this PR, maybe also assert that the `moved` attribute gets set correctly?

eldritch/test_events.py
```diff
@@ -262,6 +296,151 @@ def testMoveMultipleSpaces(self):
...
+
+  def testMoveMultipleThroughMonsterFight(self):
```
fingoldfish, line 315: Is the key difference between this test and testMoveOneSpaceToMonster that this one starts with the character moving multiple spaces? Maybe add a comment to this test.

eldritch/test_events.py
```diff
@@ -262,6 +296,151 @@ def testMoveMultipleSpaces(self):
...
+    movement.resolve(self.state, "Graveyard")
+
+    choice = self.resolve_to_choice(MultipleChoice)
+
+    self.assertEqual(choice.choices, ["Fight", "Evade"])
+    choice.resolve(self.state, "Evade")
+
+    with mock.patch.object(events.random, "randint", new=mock.MagicMock(return_value=3)):
+      next_choice = self.resolve_to_choice(MultipleChoice)
+    self.assertEqual(sorted(next_choice.choices), ["Fight", "Flee"])
```
fingoldfish, line 382: Pro tip: you can use the handy assertCountEqual and not worry about the order.

eldritch/test_events.py
```diff
@@ -262,6 +296,151 @@ def testMoveMultipleSpaces(self):
...
+    next_choice.resolve(self.state, "Flee")
+
+    with mock.patch.object(events.random, "randint", new=mock.MagicMock(return_value=5)):
+      movement = self.resolve_to_choice(CityMovement)
+
+    self.assertFalse(movement.choices)
+    movement.resolve(self.state, movement.none_choice)
+    self.resolve_until_done()
```
fingoldfish, line 390: You should add a comment here - using resolve_until_done() also tests that you don't have to evade the zombie again when you're done with your movement.

eldritch/test_events.py
```diff
@@ -262,6 +296,151 @@ def testMoveMultipleSpaces(self):
...
+    self.assertIn("Rivertown", movement.choices)
+    movement.resolve(self.state, "Rivertown")
+    choice = self.resolve_to_choice(MultipleChoice)
+    # TODO: Choose which one to evade first
+    # For now, Zombie is the first
+    self.assertEqual(choice.choices, ["Fight", "Evade"])
+    with mock.patch.object(events.random, "randint", new=mock.MagicMock(return_value=4)):
+      choice.resolve(self.state, "Evade")
+      next_choice = self.resolve_to_choice(MultipleChoice)
+      self.assertEqual(sorted(next_choice.choices), ["Fight", "Flee"])
```
fingoldfish, line 425: Same nitpick - can put the asserts outside of the patch block.

eldritch/test_events.py
```diff
@@ -262,6 +296,151 @@ def testMoveMultipleSpaces(self):
...
+
+    self.assertFalse(movement.choices)
+    movement.resolve(self.state, movement.none_choice)
+    self.assertTrue(movement.is_resolved())
+    self.assertEqual(self.char.place.name, "Easttown")
+    self.assertEqual(self.char.stamina, 4)
+    self.assertEqual(self.char.sanity, 5)
+    self.assertEqual(self.char.movement_points, 0)
+    # TODO: ^ should this pass?
+
```
fingoldfish, line 443: Let's add a TODO for one more test - if the character has a motorcycle, they should not be able to exhaust the motorcycle to get back any movement points after entering combat. I haven't thought too deeply about how to do that yet - maybe set movement points to -float("inf") instead of 0?

fingoldfish (changes requested):

eldritch/test_events.py
fingoldfish: Okay, one more TODO:
a test where you fight a dream flier in a location with a gate, lose, get pulled through the gate, immediately cast Find Gate, return, and then don't have to fight again.

eldritch/test_events.py
petercombs: Done

eldritch/monsters.py
fingoldfish: Should still be self.place.name instead of self.place

eldritch/eldritch.py
fingoldfish: Same comment still applies. I think EvadeOrFightAll probably works better, since you haven't made the same modifications to get_triggers.

eldritch/events.py
fingoldfish: This nitpick and others in events.py look like they got missed?

eldritch/test_events.py
fingoldfish: Would still like for the `moved` attribute to be tested.

fingoldfish (approved):

PR #32, opened 2021-05-25
Close Location and Tests

petercombs:


eldritch/events.py
```diff
@@ -174,9 +174,10 @@ def resolve(self, state):
       self.done = True
       return True
     assert self.dest in self.character.place.connections
-    self.character.place = self.dest
-    self.character.movement_points -= 1
-    self.character.explored = False
+    if not self.dest.closed:
```
fingoldfish, line 177: Interesting to note that if a person tries to move past a closed location, they'll get an assertionerror anyway when the next MoveOne asserts that the two locations are connected.

Not a problem, I think it's fine. The UI shouldn't let people move to or past closed locations anyway.

eldritch/events.py
```diff
@@ -1641,6 +1642,12 @@ def resolve(self, state):
       self.closed = True
       state.gates.append(state.places[self.location_name].gate)  # TODO: take a gate trophy
       state.places[self.location_name].gate = None
+      closed_until = state.places[self.location_name].closed_until or -1
+      if closed_until > state.turn_number:
+        state.event_stack.append(
+            CloseLocation(self.location_name, closed_until - state.turn_number - 1)
+        )
+        return False
```
fingoldfish, line 1650: Not keeping track of whether you've re-closed the location is slightly dangerous - if anyone were to add more code underneath the `if closed_until > state.turn_number` block, that code wouldn't be executed in the case where the location was re-closed. I would keep a boolean at the class level.

eldritch/events.py
```diff
@@ -1938,6 +1945,41 @@ def start_str(self):
...
+    place = state.places[self.location_name]
+    place.closed_until = until
+    chars_in_place = [char for char in state.characters if char.place == place]
+    monsters_in_place = [mon for mon in state.monsters if mon.place == place]
+
+    if place.closed and self.evict:
+      to_place = next(iter(place.connections))
+      for char in chars_in_place:
+        state.event_stack.append(ForceMovement(char, to_place.name))
+        return False
```
fingoldfish, line 1967: This is...strange. I guess it works in that you exit the for loop after the first iteration, so you only append one event at a time. Appending multiple events to the loop at once is disallowed, but it looks like you're not doing that.

I would feel better if this were a Sequence of ForceMovements, but I can't actually justify that with anything.

eldritch/test_events.py
```diff
@@ -822,6 +822,56 @@ def testCombatChoice(self):
 
 # TODO: add tests for going unconscious/insane during a mythos/encounter.
 
+class CloseLocationTest(EventTest):
+  def testCloseForever(self):
+    place_name = "Woods"
+    self.state.event_stack.append(CloseLocation(place_name))
+    self.resolve_until_done()
+    self.assertTrue(self.state.places[place_name].closed)
```
fingoldfish, line 830: Maybe advance one or two turns and verify that the location is still closed?

eldritch/test_events.py
```diff
@@ -822,6 +822,56 @@ def testCombatChoice(self):
...
+    monster = next(iter(self.state.monsters))
+    monster.place = place
+    self.char.place = place
+    self.state.event_stack.append(CloseLocation(place_name, 1))
+    self.resolve_until_done()
+    self.assertEqual(place.closed_until, self.state.turn_number + 2)
+    self.assertEqual(self.char.place, place)
+    self.assertEqual(monster.place, place)
+    self.assertFalse(place.closed)
+    self.state.event_stack.append(GateCloseAttempt(self.char, place_name))
```
fingoldfish, line 845: Use some blank lines to split this test up into logical chunks. e.g. setup, first round of assertions, closing the gate, second round of assertions

eldritch/test_events.py
```diff
@@ -822,6 +822,56 @@ def testCombatChoice(self):
...
+    with mock.patch.object(events.random, "randint", new=mock.MagicMock(return_value=5)):
+      fight_lore = self.resolve_to_choice(MultipleChoice)
+      self.assertEqual(fight_lore.choices, ["Close with fight", "Close with lore", "Don't close"])
```
fingoldfish, line 848: Probably best not to assert on this in case we change the strings in the future. You can just choose the first choice out of choices.

fingoldfish (approved):

eldritch/events.py
petercombs: To be clear, you're saying don't use a @property for CityPlace.closed, and instead just have it be a boolean like in my original implementation?

eldritch/events.py
petercombs: Agree, changed.

eldritch/test_events.py
petercombs: Roger

eldritch/test_events.py
```diff
@@ -822,6 +820,107 @@ def testCombatChoice(self):
...
+    self.assertEqual(monster.place, place)
+    self.assertFalse(place.closed)
+
+    self.state.event_stack.append(GateCloseAttempt(self.char, place_name))
+    with mock.patch.object(events.random, "randint", new=mock.MagicMock(return_value=5)):
+      fight_lore = self.resolve_to_choice(MultipleChoice)
+      fight_lore.resolve(self.state, fight_lore.choices[0])
+      self.resolve_until_done()
+
+    self.assertEqual(place.closed_until, self.state.turn_number + 2)
```
fingoldfish, line 862: And also assert that place.closed is True.

eldritch/test_events.py
```diff
@@ -822,6 +820,107 @@ def testCombatChoice(self):
...
+    self.state.event_stack.append(CloseLocation(place_name, for_turns=1, evict=False))
+    self.char.place = place
+    self.resolve_until_done()
+
+    while self.state.turn_phase != 'movement':
+      self.state.next_turn()
+      self.resolve_until_done()
+    self.assertTrue(place.closed)
+    self.state.event_stack.append(
+      Sequence([MoveOne(self.char, self.state.places["Downtown"]), EndMovement(self.char)], self.char)
```
fingoldfish, line 914: Note: EndMovement is no longer a thing. You may have to resolve the CityMovement event (or just manually adjust the game turn and phase).

eldritch/test_events.py
```diff
@@ -1536,5 +1561,102 @@ def testInvalidSale(self):
...
+
+  def testMoveFromClosedLocation(self):
+    place_name = "Merchant"
+    place = self.state.places[place_name]
+    self.char.place = place
+    self.advance_turn(self.state.turn_number, 'mythos')
+    self.resolve_until_done()
+    self.state.event_stack.append(CloseLocation(place_name, for_turns=1, evict=False))
+    self.resolve_until_done()
+    self.assertTrue(place.closed)
```
fingoldfish, line 1647: Assert that the character is still there?

eldritch/test_events.py
```diff
@@ -1536,5 +1561,102 @@ def testInvalidSale(self):
...
+
+    self.assertEqual(place.closed_until, self.state.turn_number + 2)
+    self.assertEqual(self.char.place.name, "Uptown")
+    self.assertEqual(monster.place.name, "Uptown")
+    self.assertTrue(place.closed)
+
+    self.advance_turn(self.state.turn_number + 1, 'movement')
+    movement = self.resolve_to_choice(CityMovement)
+    movement.resolve(self.state, movement.none_choice)
+    with mock.patch.object(events.random, "randint", new=mock.MagicMock(return_value=5)):
```
fingoldfish, line 1606: Probably worth adding a comment here that you need to kill the cultist to continue advancing the turn properly.

fingoldfish (approved):

eldritch/test_events.py
```diff
@@ -1536,5 +1561,102 @@ def testInvalidSale(self):
       self.resolve_until_done()
 
 
+class CloseLocationTest(EventTest):
+  def testCloseForever(self):
+    place_name = "Woods"
+    self.state.event_stack.append(CloseLocation(place_name))
+    self.resolve_until_done()
+    self.assertTrue(self.state.places[place_name].closed)
+    self.char.place = self.state.places["Uptown"]
```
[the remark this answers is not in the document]
petercombs: I think I meant to test that they got evicted... Changing the test to include that.

PR #33, opened 2021-05-27
First sketch of ancient ones

petercombs:


eldritch/ancient_ones.py
```diff
@@ -0,0 +1,147 @@
...
+    if isinstance(thing, Monster) and thing.name == "Cultist":
+      return {'combat_difficulty': -2, 'movement': 'flying'}.get(attribute, 0)
+    if isinstance(thing, events.GateCloseAttempt):
+      return 3
+    return 0
+
+  def attack(self, state):
+    for char in state.characters:
+      check = events.Check(char, "luck", self.luck_modifier)
+      state.event_stack.append(
```
fingoldfish, line 67: You should be appending a sequence of pass/fails, not appending all of them on top of eachother.

eldritch/ancient_ones.py
```diff
@@ -0,0 +1,147 @@
...
+      return [events.Loss(owner.character, {'stamina': 1})]
+    #TODO: Discard weather cards
+    return []
+
+  def get_modifier(self, thing, attribute):
+    if isinstance(thing, Monster) and thing.name == "Cultist":
+      return {'toughness': 2}.get(attribute, 0)
+
+  def awaken(self, state):
+    #TODO: Roll a die for each item, discard on a failure
```
fingoldfish, line 105: This makes me think that awaken should be an event. I'm pretty sure attack needs to be an event, but I'm only like 68.3% sure that awaken needs to be an event. You'll probably have to define some custom events in this file.

eldritch/ancient_ones.py
```diff
@@ -0,0 +1,147 @@
+import abc
+
+from eldritch import events, places
+from eldritch.monsters import Monster
+
+
```
petercombs, line 6: AncientOne should also(?) probably be a subclass of mythos.GlobalEffect to get all the default triggers, interrupts, and modifiers.

eldritch/events.py
```diff
@@ -1643,7 +1643,12 @@ def resolve(self, state):
       state.places[self.location_name].gate = None
 
     if self.seal_choice is None:
-      if self.character.clues < 5:  # TODO: this can also have modifiers
+      difficulty = 5
+      difficulty += state.ancient_one.get_modifier(self, 'difficulty')
```
fingoldfish, line 1647: Use state.get_modifier so that it takes global effects into account too.

eldritch/ancient_ones.py
```diff
@@ -0,0 +1,147 @@
...
+    self.name = "The Yellow King"
+    self.attributes = ['physical resistance']
+    self.combat_difficulty = None
+    self.luck_modifier = 1
+
+  def awaken(self, state):
+    self.combat_difficulty = state.terror
+
+  def get_modifier(self, thing, attribute):
+    if isinstance(thing, Monster) and thing.name == "Cultist":
```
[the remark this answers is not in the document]
petercombs: That was my first instinct, but Cultist is a function not a class. Also, which monsters are those?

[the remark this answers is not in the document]
fingoldfish: Oh, I forgot Cultist is still a function. Eventually, most monsters will be their own class, since a lot of them need to interrupt combat in some way for their special text to take effect. I would recommend turning Cultist into its own class now and having `__init__` just call the parent `__init__` with the correct default args (similar to the Bullwhip).
See also: http://www.arkhamhorrorwiki.com/Child_of_the_Goat

petercombs:
@fingoldfish I have no idea how we want to write tests for the ancient ones... if you want to sketch some out, I can copy the pattern, either in this PR or a new one.

eldritch/ancient_ones.py
fingoldfish: Did you mean to go back and fix this?

eldritch/ancient_ones.py
```diff
@@ -0,0 +1,169 @@
+import abc
+
+from eldritch import events, places, characters, monsters, values
+
+
+class AncientOne(metaclass=abc.ABCMeta):
+  # pylint: disable=unused-argument
+  def __init__(self):
```
fingoldfish, line 8: Looks like all ancient ones need a name, a max_doom, and an attributes. Let's make them part of the `__init__` method to makes sure they always get set.

eldritch/ancient_ones.py
```diff
@@ -0,0 +1,169 @@
...
+    self.doom = 0
+    self.combat_difficulty = 0
```
fingoldfish, line 10: You don't use combat_difficulty anywhere - it should get used in get_modifier (i.e. when thing == self).

eldritch/ancient_ones.py
```diff
@@ -0,0 +1,169 @@
...
+    self.attributes = []
+
+  def get_interrupts(self, owner, state):
```
fingoldfish, line 13: This should be named get_interrupt, though you should also be inheriting it from GlobalEffect.

eldritch/ancient_ones.py
```diff
@@ -0,0 +1,169 @@
...
+    super().__init__()
+    self.name = "SquidFace"
+    self.max_doom = 13
+    self.combat_difficulty = -6
+
+  def get_modifier(self, thing, attribute):
+    if isinstance(thing, monsters.Cultist):
+      return {"horror": -2, "horror_damage": 2}.get(attribute, 0)
+    if isinstance(thing, characters.Character):
+      return {"maximum_sanity": -1, "maximum_stamina": -1}
```
fingoldfish, line 46: You forgot the `.get(attribute, 0)` here

eldritch/ancient_ones.py
```diff
@@ -0,0 +1,169 @@
...
+
+class BlackPharaoh(AncientOne):
+  def __init__(self):
+    super().__init__()
+    self.max_doom = 11
+    self.attributes = ["magical resistance"]
+    self.combat_difficulty = -4
+    self.lore_modifier = 1
+
+  def get_modifier(self, thing, attribute):
```
fingoldfish, line 146: This should be get_override. get_modifier is used for numeric values; get_override is used for overriding a specific attribute.

eldritch/ancient_ones.py
```diff
@@ -0,0 +1,169 @@
...
+
+class Wendigo(AncientOne):
+  def __init__(self):
+    super().__init__()
+    self.name = "Wendigo"
+    self.max_doom = 11
+    self.combat_difficulty = -3
+    self.fight_modifier = 1
+
+  def get_interrupts(self, owner, state):
```
fingoldfish, line 110: interrupts are events that happen before the indicated event. triggers are events that happen after the indicated event. In this case, you want to use a trigger to indicate that this happens after the movement is finished, not before the movement starts.

Also, pretty sure you want the function signature to be `get_trigger(self, event, state)` as opposed to `owner`

eldritch/events.py
```diff
@@ -2619,11 +2619,17 @@ def resolve(self, state):
         return False
 
     if self.seal_choice is None:
-      if self.character.clues < 5:  # TODO: this can also have modifiers
+      difficulty = 5
+      difficulty += state.ancient_one.get_modifier(self, "difficulty")
```
fingoldfish, line 2623: Let's not name this difficulty - there is already a difficulty for the gate close attempt fight/lore check.
Maybe clues? seal_clues? Something like that.

eldritch/events.py
```diff
@@ -2619,11 +2619,17 @@ def resolve(self, state):
...
+      difficulty += state.get_modifier(self, "difficulty")
```
fingoldfish, line 2624: You have double-counted here a bit - when you call `state.get_modifier`, it will get the modifier from all globals. The globals include: the environment, the rumor, the ancient one, and any other global effects. You do not need to list them individually here.

eldritch/monsters.py
```diff
@@ -97,6 +97,11 @@ def has_attribute(self, attribute, state, char):
       return state_override
     return attribute in self._attributes
 
+  def get_movement(self, state):
+    ancient_modifier = state.ancient_one.get_modifier(self, "movement")
```
fingoldfish, line 101: Let's use get_override instead of get_modifier here (get_modifier is for numeric values). See also the thing above about using the state's get_override instead of using get_modifier.

eldritch/monsters.py
```diff
@@ -208,8 +214,9 @@ def Hound():
   )
 
 
-def Maniac():  # TODO: custom class when we add globals
-  return Monster("Maniac", "normal", "moon", {"evade": -1, "combat": 1}, {"combat": 1}, 1)
+class Maniac(Monster):  # TODO: custom class when we add globals
```
fingoldfish, line 217: Maybe change the TODO to reflect what's missing - the combat difficulty and damage need to change based on the state.

eldritch/ancient_ones.py
```diff
@@ -0,0 +1,169 @@
...
+  def attack(self, state):
+    self.doom = min(self.doom + 1, self.max_doom)
+    # TODO: Each character lowers max sanity or stamina
+
+
+class YellowKing(AncientOne):
+  def __init__(self):
+    super().__init__()
+    self.name = "The Yellow King"
+    self.attributes = ["physical resistance"]
```
fingoldfish, line 58: Attributes should probably be a set instead of a list.

eldritch/ancient_ones.py
```diff
@@ -0,0 +1,169 @@
...
+
+  def get_modifier(self, thing, attribute):
+    if isinstance(thing, monsters.Cultist):
+      return {"horror": -2, "horror_damage": 2}.get(attribute, 0)
+    if isinstance(thing, characters.Character):
+      return {"maximum_sanity": -1, "maximum_stamina": -1}
+    return 0
+
+  def attack(self, state):
+    self.doom = min(self.doom + 1, self.max_doom)
```
fingoldfish, line 50: The more I look at this, the more I think that the attack needs to be an event, and not just some code. For example, the redeemed cultist has an ability that needs to trigger when a doom token is added to the doom track.

eldritch/ancient_ones.py
```diff
@@ -0,0 +1,169 @@
...
+
+  def attack(self, state):
+    checks = []
+    for char in state.characters:
+      check = events.Check(char, "luck", self.luck_modifier)
+      checks.append(
+          events.PassFail(char, check, events.Nothing(), events.Loss(char, {"sanity": 2}))
+      )
+    state.event_stack.append(events.Sequence(checks))
+    self.luck_modifier -= 1
```
fingoldfish, line 81: Same thing - I definitely think the attack should return an event, and not execute any code. You'll notice there is a subtle bug here - you have changed the luck modifier before executing any of the luck checks added to the event stack instead of after.

eldritch/ancient_ones.py
```diff
@@ -0,0 +1,169 @@
...
+    self.combat_difficulty = -4
+    self.lore_modifier = 1
+
+  def get_modifier(self, thing, attribute):
+    if isinstance(thing, monsters.Cultist):
+      return {"attribute": "endless"}.get(attribute, 0)
+    return 0
+
+  def awaken(self, state):
+    for char in state.characters:
```
fingoldfish, line 152: Similar to attack, awaken should probably also return an event.

eldritch/ancient_ones.py
```diff
@@ -0,0 +1,169 @@
...
+  def __init__(self):
+    super().__init__()
+    self.max_doom = 11
+    self.attributes = ["magical resistance"]
+    self.combat_difficulty = -4
+    self.lore_modifier = 1
+
+  def get_modifier(self, thing, attribute):
+    if isinstance(thing, monsters.Cultist):
+      return {"attribute": "endless"}.get(attribute, 0)
```
fingoldfish, line 148: Let's add some tests. Probably a new file where you create a game, set up the ancient one, and then validate that attributes on various monsters get set correctly.

eldritch/ancient_ones.py
```diff
@@ -0,0 +1,169 @@
...
+    self.name = "Wendigo"
+    self.max_doom = 11
+    self.combat_difficulty = -3
+    self.fight_modifier = 1
+
+  def get_interrupts(self, owner, state):
+    if (
+        isinstance(owner, events.CityMovement)
+        and owner.is_resolved()
+        and isinstance(owner.character.place, places.Street)
```
fingoldfish, line 114: This should also have a test.

eldritch/ancient_ones.py
```diff
@@ -0,0 +1,169 @@
...
+    self.luck_modifier = 1
+    self.max_doom = 13
+
+  def awaken(self, state):
+    self.combat_difficulty = state.terror
+
+  def get_modifier(self, thing, attribute):
+    if isinstance(thing, monsters.Cultist):
+      return {"combat_difficulty": -2, "movement": "flying"}.get(attribute, 0)
+    if isinstance(thing, events.GateCloseAttempt):
```
fingoldfish, line 69: You should probably check the attribute here too.

fingoldfish (changes requested):

eldritch/ancient_ones.py
fingoldfish: To be more specific, probably a test that inherits from EventTest, has a character in the street, and then advance to the mythos phase and then the next upkeep. Validate that the character loses a stamina.
Note that this ancient one's ability applies to the mythos phase, not the movement phase. So you'll want to add a test to make sure a character that is in a location during the movement phase but gets kicked out during the encounter phase still loses one stamina.
Lastly, you'll want to have a negative test that validates that a character in a location does not lose one stamina.

eldritch/ancient_ones.py
```diff
@@ -0,0 +1,147 @@
...
+    return []
+
+  @abc.abstractmethod
+  def attack(self, state):
+    raise NotImplementedError
+
+
+class SquidFace(AncientOne):
+  def __init__(self):
+    super(AncientOne, self).__init__()
```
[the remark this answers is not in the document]
petercombs: Now using just `super().__init__()`

eldritch/ancient_ones.py
```diff
@@ -0,0 +1,147 @@
...
+    # TODO: Each character lowers max sanity or stamina
+
+
+class YellowKing(AncientOne):
+  def __init__(self):
+    super(AncientOne, self).__init__()
+    self.name = "The Yellow King"
+    self.attributes = ['physical resistance']
+    self.combat_difficulty = None
+    self.luck_modifier = 1
```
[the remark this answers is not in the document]
petercombs: Oversight. New init method resolves that.

eldritch/ancient_ones.py
```diff
@@ -0,0 +1,147 @@
...
+    self.attributes = []
+
+  def get_interrupts(self, owner, state):
+    return []
+
+  def get_modifier(self, thing, attribute):
+    return 0
+
+  def awaken(self, state):
+    return []
```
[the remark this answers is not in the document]
petercombs: Now using AncientOneAwaken.

eldritch/ancient_ones.py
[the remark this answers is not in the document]
petercombs: Done.

eldritch/ancient_ones.py
petercombs: Leaving the to-do for now, but I am using the AncientOneAwaken event.

eldritch/events.py
petercombs: Done.

eldritch/ancient_ones.py
petercombs: Done

eldritch/ancient_ones.py
petercombs: Fixed.

eldritch/ancient_ones.py
petercombs: Fixed!

eldritch/monsters.py
petercombs: Done!

eldritch/events.py
petercombs: Done.

eldritch/events.py
petercombs: Done.

eldritch/ancient_ones.py
petercombs: Done.

eldritch/ancient_ones.py
petercombs: Done.

eldritch/ancient_ones.py
petercombs: Done.

eldritch/ancient_ones.py
petercombs: Done.

eldritch/ancient_ones.py
petercombs: Done.

eldritch/ancient_ones.py
petercombs: I put some todo's in the obvious places (here and in OpenGate) to make an AddDoom event. Leaving the doom modifier here for now.

eldritch/ancient_ones.py
petercombs: Done.

eldritch/monsters.py
petercombs: I think this no longer needs a TODO...

eldritch/ancient_ones.py
petercombs: I created a few tests---can add some more on all of the ancient ones if you like how they're set up.

eldritch/ancient_ones.py
petercombs: Done.

eldritch/ancient_ones.py
petercombs: Done.

petercombs:
@fingoldfish Added some tests---if you like the way I'm going with 'em and want me to add more (i.e. at least one for each of the AOs), I can.

eldritch/ancient_ones.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+class BlackPharaoh(AncientOne):
+  def __init__(self):
+    super().__init__("The Thousand Masks", 11, {"magical resistance"}, -4)
+    self.lore_modifier = 1
+    # TODO: Add masks to monster cup
+
+  def get_override(self, thing, attribute):
+    if attribute == "endless":
```
fingoldfish, line 152: Pretty sure this is supposed to only apply to cultists.

eldritch/ancient_ones.py
```diff
@@ -0,0 +1,181 @@
...
+      has_clues = values.AttributePrerequisite(char, "clues", "at least", 1)
+      checks.append(
+          events.Sequence([
+              events.PassFail(char, check, events.Nothing(), events.Loss(char, {"clues": 1})),
+              events.PassFail(char, has_clues, events.Nothing(), events.Nothing())
+              # TODO: Devoured if no clue tokens
+          ], char)
+
+      )
+    self.lore_modifier -= 1
```
fingoldfish, line 180: Same thing.

eldritch/test_ancient_ones.py
```diff
@@ -0,0 +1,66 @@
...
+    self.assertEqual(self.state.get_modifier(monsters.Cultist(), "horror_damage"), 2)
+
+  def testYellowKingMonsters(self):
+    self.state.ancient_one = ancient_ones.YellowKing()
+    cultist = next(
+        monster for monster in self.state.monsters if isinstance(monster, monsters.Cultist)
+    )
+    self.assertTrue(cultist.has_attribute("flying", self.state, self.char))
+
+
```
fingoldfish, line 64: Let's add some TODOs to test the other things that haven't been tested yet (e.g. combat ratings of cultists for the yellow king, wendingo cancelling an environment, etc.). Or you could just add the tests in this pr - either way.

fingoldfish (changes requested):

eldritch/ancient_ones.py
```diff
@@ -0,0 +1,181 @@
...
+class AncientOne(mythos.GlobalEffect, metaclass=abc.ABCMeta):
+  # pylint: disable=unused-argument
+  def __init__(self, name, max_doom, attributes, combat_rating):
+    self.name = name
+    self.max_doom = max_doom
+    self.doom = 0
+    self.combat_rating = combat_rating
+    self.attributes = attributes
+
+  def get_modifier(self, thing, attribute):
```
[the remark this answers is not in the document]
petercombs: I can't think of anything that would affect the AO's combat rating (e.g. items, abilities, etc), so maybe it doesn't need to be a function at all, and we can just ask for `ancient_one.combat_rating`...

eldritch/events.py
```diff
@@ -560,9 +560,15 @@ def resolve(self, state):
       new_val = old_val + adjustment
       new_val = max(new_val, 0)
       if attr == "stamina":
-        new_val = min(new_val, self.character.max_stamina)
+        new_val = min(
```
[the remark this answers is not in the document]
petercombs: Done.

eldritch/events.py
```diff
@@ -2692,11 +2698,13 @@ def resolve(self, state):
         return
 
     if self.seal_choice is None:
-      if self.character.clues < 5:  # TODO: this can also have modifiers
+      seal_clues = 5
+      seal_clues += state.get_modifier(self, "seal_clues")
```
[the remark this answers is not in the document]
petercombs: 💩 happens. Not that difficult.

eldritch/ancient_ones.py
```diff
@@ -0,0 +1,181 @@
...
+  def __init__(self):
+    super().__init__("Dummy", 10, set(), 0)
+
+  def attack(self, state):
+    return AncientOneAttack([])
+
+
+class SquidFace(AncientOne):
+  def __init__(self):
+    super().__init__("SquidFace", 13, set(), -6)
```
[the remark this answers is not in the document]
petercombs: Done.

eldritch/ancient_ones.py
```diff
@@ -0,0 +1,181 @@
...
+    super().__init__("The Yellow King", 13, {"physical resistance"}, None)
+    self.luck_modifier = 1
+
+  def awaken(self, state):
+    self.combat_rating = state.terror
+    return AncientOneAwaken([])
+
+  def get_modifier(self, thing, attribute):
+    if isinstance(thing, monsters.Cultist):
+      return {"combat_difficulty": -2}.get(attribute, 0)
```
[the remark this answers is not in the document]
petercombs: Oops.

eldritch/ancient_ones.py
```diff
@@ -0,0 +1,181 @@
...
+    return AncientOneAwaken([])
+
+  def attack(self, state):
+    checks = []
+    for char in state.characters:
+      check = events.Check(char, "fight", self.fight_modifier)
+      checks.append(
+          events.PassFail(char, check, events.Nothing(), events.Loss(char, {"stamina": 2}))
+      )
+    self.fight_modifier -= 1
```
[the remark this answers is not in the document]
petercombs: Same as what? I think before there was maybe some sort of race condition where the modifier was getting changed prior to the check actually happening, but now when I create the Check() it should be using the old value as the threshold.

eldritch/test_ancient_ones.py
petercombs: Done

fingoldfish (changes requested):
Just a few more things to take care of.

eldritch/ancient_ones.py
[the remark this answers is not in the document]
petercombs: TODO added in upcoming commit.

eldritch/ancient_ones.py
```diff
@@ -0,0 +1,179 @@
...
+
+  def get_override(self, thing, attribute):
+    if isinstance(thing, monsters.Cultist) and attribute == "endless":
+      return True
+    return super().get_override(thing, attribute)
+
+  def awaken(self, state):
+    checks = []
+    for char in state.characters:
+      has_clues = values.AttributePrerequisite(char, "clues", "at least", 1)
```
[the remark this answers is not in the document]
petercombs: Fix in upcoming commit.

eldritch/ancient_ones.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+class Wendigo(AncientOne):
+  def __init__(self):
+    super().__init__("Wendigo", 11, set(), -3)
+    self.fight_modifier = 1
+
+  def get_interrupt(self, event, state):
+    # TODO: Discard weather cards
+    if isinstance(event, events.ActivateEnvironment) and event.environment_type == "weather":
```
fingoldfish, line 110: Should be event.env.environment_type

eldritch/test_ancient_ones.py
```diff
@@ -0,0 +1,92 @@
...
+    self.char.sanity = 3
+    self.state.event_stack.append(events.Gain(self.char, {"sanity": 5, "stamina": 5}))
+    self.resolve_until_done()
+    self.assertEqual(self.char.stamina, 4)
+    self.assertEqual(self.char.sanity, 4)
+
+  def testSquidFaceMonsters(self):
+    self.assertEqual(self.state.get_modifier(monsters.Cultist(), "horror_damage"), 0)
+    self.state.ancient_one = ancient_ones.SquidFace()
+    self.assertEqual(self.state.get_modifier(monsters.Cultist(), "horror"), -2)
```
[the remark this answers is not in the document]
petercombs: Done.

fingoldfish (approved):

PR #98, opened 2022-01-24
Bind Monster

petercombs:


eldritch/events.py
```diff
@@ -2500,8 +2503,9 @@ def choice_count(self):
 
 class CombatChoice(ItemChoice):
 
-  def __init__(self, character, prompt):
+  def __init__(self, character, prompt, combat_round=None):
```
petercombs, line 2506: I'm not in love with this as a feature. It's optional because there are a couple encounters that directly create CombatChoice's without a related CombatRound.

eldritch/items.py
```diff
@@ -238,8 +239,41 @@ def DreadCurse(idx):
...
+  def __init__(self, idx):
+    super().__init__(
+        "Bind Monster", idx, {"magical": float("inf"), "physical": float("inf")}, 2,
+        4, 2,
+    )
+    self.combat_round = None
+
+  def get_required_successes(self, state):  # pylint: disable=unused-argument
+    self.combat_round = state.event_stack[
+        -3
```
petercombs, line 252: Is there some better way for the spell to get access to the monster that we're facing?

eldritch/items.py
```diff
@@ -238,8 +239,41 @@ def DreadCurse(idx):
...
+    ):
+      ret = super().get_usable_interrupt(event, owner, state)
+      return ret
+    return None
 
+  def get_cast_event(self, owner, state):
+    return events.Sequence(
+        [
+            events.PassCombatRound(self.combat_round),
+            events.TakeTrophy(owner, self.combat_round.monster),
```
petercombs, line 268: With a LootPinata, is there a risk that we TakeTrophy on it twice?

eldritch/items.py
fingoldfish: The monster is also an attribute of the CombatChoice. Though as you pointed out, there are some combats that don't have a monster, but I think you would handle that in get_usable_interrupt instead. So I might simplify this to something like
```
choice = state.event_stack[-2]
return choice.monster.toughness(state, choice.character)
```

eldritch/items.py
```diff
@@ -238,8 +239,41 @@ def DreadCurse(idx):
...
+    self.combat_round = None
+
+  def get_required_successes(self, state):  # pylint: disable=unused-argument
+    self.combat_round = state.event_stack[
+        -3
+    ]  # CombatRound[-3] > CombatChoice[-2] > CastSpell[-1]
+    return self.combat_round.monster.toughness(state, self.combat_round.character)
+
+  def get_usable_interrupt(self, event, owner, state):
+    if isinstance(event, events.CombatChoice) and isinstance(
```
fingoldfish, line 257: Might just update the CombatChoice constructor to always take a monster as an argument (even if it's None) to make sure that the monster is always set on the CombatChoice. I'm honestly not sure why I didn't do that before.

eldritch/items.py
fingoldfish: Here's what I would do to make it so that TakeTrophy isn't done twice - I would actually change PassCombatRound to include the TakeTrophy and overwhelming damage, and then update CombatRound to just stick a PassCombatRound on the stack. (And of course change the cast event to just be PassCombatRound + Discard.)

I also recommend against saving combat_round as a property on the spell, mostly because it may lead to other subtle bugs if something changes in the CastSpell event.

eldritch/items.py
```diff
@@ -238,8 +239,41 @@ def DreadCurse(idx):
...
+            events.PassCombatRound(self.combat_round),
+            events.TakeTrophy(owner, self.combat_round.monster),
+            # In case we go insane, take trophy explicitly
+            events.DiscardSpecific(owner, [self])
+        ],
+        owner,
+    )
+
+
+class EnchantWeapon(CombatSpell):
```
fingoldfish, line 276: If you add an extra blank line after this class definition, I think git diff will figure out that you didn't change it. The diff algorithm is really bad.

eldritch/items.py
```diff
@@ -238,8 +239,41 @@ def DreadCurse(idx):
...
+    self.combat_round = state.event_stack[
+        -3
+    ]  # CombatRound[-3] > CombatChoice[-2] > CastSpell[-1]
+    return self.combat_round.monster.toughness(state, self.combat_round.character)
+
+  def get_usable_interrupt(self, event, owner, state):
+    if isinstance(event, events.CombatChoice) and isinstance(
+        event.combat_round.monster, monsters.Monster
+    ):
+      ret = super().get_usable_interrupt(event, owner, state)
```
fingoldfish, line 260: Can simplify by eliminating ret.

eldritch/events.py
```diff
@@ -2983,6 +2989,27 @@ def finish_str(self):
...
+  def resolve(self, state):
+    self.combat_round.defeated = True
+    self.done = True
+    return True
+
+  def is_resolved(self) -> bool:
+    return self.done
+
+  def start_str(self):
+    return self.log_message
```
fingoldfish, line 3007: I might (a) change the log message to look more like the others (e.g. character.name passed a combat round against a monster.name) and (b) put this in finish_str instead of start_str

eldritch/test_combat.py
```diff
@@ -1978,5 +1978,108 @@ def testDeactivatesOnCombatLoss(self):
     self.assertTrue(witch.has_attribute("magical resistance", self.state, self.char))
 
 
+class BindMonsterTest(EventTest):
```
fingoldfish, line 1981: In addition to regular testing, can I ask that you add BindMonster to a character's possessions for the encounter at the bank and the encounter at the hospital? We should verify that BindMonster is not usable in either of those combats.

eldritch/test_combat.py
```diff
@@ -1978,5 +1978,108 @@ def testDeactivatesOnCombatLoss(self):
...
+    worm = monsters.GiantWorm()
+    self.start(worm)
+    self.assertEqual(self.char.sanity, 3)
+    self.assertEqual(self.char.stamina, 5)
+    self.state.event_stack.append(self.state.usables[0]["Bind Monster0"])
+    choose_ignore = self.resolve_to_choice(SpendChoice)
+    self.spend("sanity", 2, choose_ignore)
+    choice = self.resolve_to_choice(SpendChoice)
+    choice.resolve(self.state, "Cast")
+    with mock.patch.object(
```
fingoldfish, line 2012: Can this fit on one line?

eldritch/test_combat.py
```diff
@@ -1978,5 +1978,108 @@ def testDeactivatesOnCombatLoss(self):
...
+        events.random, "randint", new=mock.MagicMock(return_value=5)
+    ):
+      choice = self.resolve_to_choice(CombatChoice)
```
fingoldfish, line 2015: Oh, this is kind of weird. You successfully cast the spell, but then you have to resolve the combat choice? I think you want the spell to cancel both the CombatChoice and the CombatRound, (otherwise, you'll end up making a combat check again, and be asked if you want to spend clue tokens, etc.).

eldritch/test_combat.py
```diff
@@ -1978,5 +1978,108 @@ def testDeactivatesOnCombatLoss(self):
...
+
+  def testSanePass(self):
+    self.char.sanity = 4
+    self.char.fight_will_slider = 0
+    worm = monsters.GiantWorm()
+    self.start(worm)
+    self.assertEqual(self.char.sanity, 3)
+    self.assertEqual(self.char.stamina, 5)
+    self.state.event_stack.append(self.state.usables[0]["Bind Monster0"])
+    choose_ignore = self.resolve_to_choice(SpendChoice)
```
fingoldfish, line 2008: It's a time honored tradition to copy-paste other working tests, but you should probably rename this from `choose_ignore` to something like `cast_choice`. Same for tests below.

eldritch/test_combat.py
```diff
@@ -1978,5 +1978,108 @@ def testDeactivatesOnCombatLoss(self):
...
+    # self.assertEqual(self.char.stamina, 4) # Overwhelming 1
+    # TODO: Overwhelming should also happen
+
+  def testSaneFail(self):
+    self.char.sanity = 3
+    self.char.fight_will_slider = 0
+    self.char.possessions.append(items.TommyGun(0))
+    cultist = monsters.Cultist()
+    self.start(cultist)
+    self.assertEqual(self.char.sanity, 3)  # Lost one to nightmarish
```
fingoldfish, line 2055: Comment needs updating.

eldritch/test_combat.py
```diff
@@ -1978,5 +1978,108 @@ def testDeactivatesOnCombatLoss(self):
...
+    self.char.sanity = 3
+    self.char.fight_will_slider = 0
+    worm = monsters.GiantWorm()
+    self.start(worm)
+    self.assertEqual(self.char.sanity, 2)  # Lost one to nightmarish
+    self.assertEqual(self.char.stamina, 5)
+    self.state.event_stack.append(self.state.usables[0]["Bind Monster0"])
+    choose_ignore = self.resolve_to_choice(SpendChoice)
+    self.spend("sanity", 2, choose_ignore)
+    choice = self.resolve_to_choice(SpendChoice)
```
fingoldfish, line 2034: I think you shouldn't need the second `resolve_to_choice(SpendChoice)` anymore. I think `self.spend` takes care of that for you.

eldritch/test_combat.py
```diff
@@ -1978,5 +1978,108 @@ def testDeactivatesOnCombatLoss(self):
...
+    choice.resolve(self.state, "Cast")
+    with mock.patch.object(
+        events.random, "randint", new=mock.MagicMock(return_value=4)
+    ):
+      choice = self.resolve_to_choice(CombatChoice)
+
+    with self.assertRaises(AssertionError):
+      # We already used our hands
+      choice.resolve(self.state, "Tommy Gun0")
+
```
fingoldfish, line 2069: There's a few more assertions that we should make here:
(a) after we failed to cast the spell, it should no longer be usable
(b) after one round of combat, where we presumably roll all 1s and fail to defeat the cultist (...hypothetically), we should be able to deactivate the spell and use our hands again.

fingoldfish (changes requested):

eldritch/items.py
petercombs: Ah, on attempts to implement this, it turns out that I use self.combat_round in the get_cast_event, so I think I'll just stick with only one dive into the event_stack.

eldritch/test_combat.py
petercombs: Done.

eldritch/items.py
```diff
@@ -238,8 +239,36 @@ def DreadCurse(idx):
...
 
 
-class EnchantWeapon(CombatSpell):
+class BindMonster(CombatSpell):
+  def __init__(self, idx):
+    super().__init__("Bind Monster", idx, {}, 2, 4, 2, )
+    self.combat_round = None
+
+  def get_required_successes(self, state):
+    self.combat_round = state.event_stack[-3]
```
fingoldfish, line 248: A minor nitpick here. I would recommend setting combat_round equal to `state.event_stack[-2].combat_round`. In get_usable_interrupt, you correctly validate that this is a combat choice and that the combat choice has a combat round, so I would make self.combat_round point to the object that you know exists and is a combat round (maybe the event stack will look different in the future?).

fingoldfish (approved):

PR #114, opened 2022-03-16
Heal spell

petercombs:


eldritch/test_usables.py
```diff
@@ -374,6 +374,103 @@ def testCannotBeCastOutsideOtherworldMovement(self):
...
+      self.resolve_until_done()
+
+    self.assertEqual(self.char.stamina, 3)
+    self.assertEqual(self.char.sanity, 4)
+    self.assertTrue(self.char.possessions[0].exhausted)
+
+  def testMultipleOptions(self):
+    nun = characters.CreateCharacters()["Nun"]
+    nun.stamina = 1
+    nun.place = self.char.place
```
fingoldfish, line 449: As a negative assertion, you could create another character and put them somewhere else. You don't even have to add new assertions - the assertion on the choices should be enough.

fingoldfish (approved):

PR #138, opened 2022-05-22
Automatically fix code style issues with autopep8 and commit

petercombs:


.github/workflows/python.yml
```diff
@@ -22,9 +27,13 @@ jobs:
         python -m pip install --upgrade pip
         pip install pylint autopep8
         if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
-    - name: Check formatting with autopep8
-      run: |
-        autopep8 --diff --exit-code **/*.py
+    - name: Auto-formatting with autopep8
+      uses: wearerequired/lint-action@v2
+      with:
+        auto_fix: ${{ github.event_name == 'pull_request' }}
```
fingoldfish, line 33: Will this result in a failed check without an auto fix if a commit somehow ends up on the master branch that doesn't conform to autopep8?

fingoldfish, line 33: Will this result in a failed check without an auto fix if a commit somehow ends up on the master branch that doesn't conform to autopep8?

fingoldfish (approved):

PR #139, opened 2022-05-22
DONT MERGE ME: Ensure that autopep8 will actually make a commit

petercombs:


petercombs:
This should just demonstrate that the wearerequired auto-lint action actually creates its own commit.

petercombs:
Autopep8 won't fix the quoting issue that pylint catches. I don't think there's a way around this, unless we want to try using https://github.com/Zoynels/autopep8_quotes

PR #151, opened 2022-06-18
Make items a sub-module to keep the large list of things relatively separate

petercombs:
As I start to add more unique items, the single `items.py` is getting harder to navigate.


eldritch/items/__init__.py
```diff
@@ -0,0 +1,70 @@
+from eldritch.assets import Deputy
+from .base import Item, Weapon, OneshotWeapon, Tome
+from .common import (
+    AncientTome, Automatic45, Axe, DarkCloak,
+    Derringer18, Revolver38, Dynamite,
+    TommyGun, Food, ResearchMaterials,
+    Rifle, Shotgun, CavalrySaber, Whiskey, Knife, Lantern,
+    Bullwhip, Cross)
+
+from .unique import (
```
fingoldfish, line 10: My personal request for keeping this modular is to add a `__all__` to unique (and common, spells, and base) so that you can just say `from .unique import *` here. Though I suppose not deputy, as that's a special case.

eldritch/items/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+    HolyWater, EnchantedBlade, EnchantedKnife, MagicLamp, MagicPowder, SwordOfGlory
+)
+from .spells import (BindMonster, DreadCurse, EnchantWeapon,
+                     FindGate, FleshWard, Heal, Mists, RedSign,
+                     Shrivelling, Voice, Wither)
+from .deputy import PatrolWagon, DeputysRevolver
+
+
+def CreateCommon():
```
fingoldfish, line 19: Similarly, I recommend moving this to `common.py` - makes it so that if you add a new item, there's just fewer files to update.

eldritch/items/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+    AncientTome, Automatic45, Axe, DarkCloak,
+    Derringer18, Revolver38, Dynamite,
+    TommyGun, Food, ResearchMaterials,
+    Rifle, Shotgun, CavalrySaber, Whiskey, Knife, Lantern,
+    Bullwhip, Cross)
+
+from .unique import (
+    HolyWater, EnchantedBlade, EnchantedKnife, MagicLamp, MagicPowder, SwordOfGlory
+)
+from .spells import (BindMonster, DreadCurse, EnchantWeapon,
```
fingoldfish, line 13: Food for thought: do you want to include skills/allies in the items submodule? I could go either way. On the one hand, having a single pattern for defining these cards is useful. On the other hand, `items` would become a misnomer, and `player_acquirable_cards` doesn't exactly roll off the tongue.

fingoldfish (changes requested):
Minor nits.

eldritch/items/__init__.py
petercombs: I think you could call them `trade_items` like they do [in the wiki](http://www.arkhamhorrorwiki.com/Trade_items), but also ["item"](http://www.arkhamhorrorwiki.com/Item) seems clear enough.

eldritch/items/__init__.py
petercombs: I've been told to shun `from x import *` in general, but I think here it really is the best way.

fingoldfish: Yeah, it's even discouraged by the python official documentation. But I do think it's warranted here.

fingoldfish (approved):

fingoldfish:
I am going to attempt to close and re-open this PR to see if that will rerun the checks.

PR #152, opened 2022-06-18
Unique item tranche

petercombs:


eldritch/events.py
```diff
@@ -1256,6 +1256,8 @@ def finish_str(self):
 
 def Draw(character, deck, draw_count, prompt="Choose a card", keep_count=1, target_type=None):
   cards = DrawItems(character, deck, draw_count, target_type=target_type)
+  if keep_count == 'all':
```
fingoldfish, line 1259: Do we want to support this? Better question: in what circumstances do you imagine this would be used? I could imagine this being deceptive when combined with someone like the archaeologist or salesman. Or maybe it's not deceptive because it would do exactly the right thing?

petercombs: In the expansions, there are encounters like: 
> Pass a [Speed (-1) check](http://www.arkhamhorrorwiki.com/Speed_check) to draw [Unique items](http://www.arkhamhorrorwiki.com/Unique_item) equal to your successes, keeping any [Tomes](http://www.arkhamhorrorwiki.com/Tome) that you draw and discarding the rest. ([Celeano](http://www.arkhamhorrorwiki.com/Celeano_(encounters)))

or 

> Make a [Luck (-2) check](http://www.arkhamhorrorwiki.com/Luck_check). If you succeed, draw 3 [Unique Item](http://www.arkhamhorrorwiki.com/Unique_Item) cards and keep any that cost $4 or less. Discard the rest. ([Curiositie Shoppe](http://www.arkhamhorrorwiki.com/Curiositie_Shoppe))

petercombs: I believe how this would work with the Archaeologist is that you would pass in `keep_count="all"`, but the archaeology ability would replace the original `DrawItems`  with a `DrawItems` that has one higher for DrawCount.  However, since `keep_count` is being set in the `Draw` function, they would only get to keep the original number.  We would, additionally, need to add something like a `keep_criteria` kwarg to KeepDrawn to check that the cards pass (e.g. `keep_criteria=lambda c: isinstance(c, Tome)` or `keep_criteria=lambda c: c.cost <= 4`)

fingoldfish: Alternate to keep_criteria: create a FilterDrawn event. Use a Sequence([DrawItems, FilterDrawn, KeepDrawn]), where FilterDrawn exposes the same interface as DrawItems. I don't know yet if I like that idea better; it depends on how messy KeepDrawn is going to become.

eldritch/items/unique.py
```diff
@@ -56,3 +58,228 @@ def get_trigger(self, event, owner, state):
...
+  def __init__(self, idx):
+    super().__init__("Obsidian Statue", idx, "unique", {}, {}, None, 4)
+
+  def get_usable_interrupt(self, event, owner, state):
+    if not isinstance(event, events.GainOrLoss) or not event.character == owner:
+      return None
+
+    types = [loss_type for loss_type in ("sanity", "stamina") if loss_type in event.losses]
+    seq = [events.DiscardSpecific(owner, [self])]
+    if len(types) == 1:
```
fingoldfish, line 234: Don't forget about losses of clues, where this item isn't usable.

eldritch/items/unique.py
```diff
@@ -56,3 +58,228 @@ def get_trigger(self, event, owner, state):
...
+      loss_type = types[0]
+      seq.append(events.LossPrevention(self, event, loss_type, event.losses[loss_type]))
```
fingoldfish, line 236: I might just use float("inf") here, in case something else modifies the loss.

eldritch/items/unique.py
```diff
@@ -56,3 +58,228 @@ def get_trigger(self, event, owner, state):
...
+  def __init__(self, idx):
+    super().__init__("Healing Stone", idx, "unique", {}, {}, None, 8)
+
+  # Copied get_usable{_trigger,_interrupt,} from Physician ability
+  def get_usable_trigger(self, event, owner, state):
+    if not isinstance(event, events.UpkeepActions):
+      return None
+    return self.get_usable(event, owner, state)
+
+  def get_usable_interrupt(self, event, owner, state):
```
fingoldfish, line 187: I may change these to only be usable triggers on UpkeepActions. Otherwise, it can prompt the user too many times.

eldritch/items/unique.py
```diff
@@ -56,3 +58,228 @@ def get_trigger(self, event, owner, state):
...
+    else:
+      gain = events.BinaryChoice(
+          owner,
+          "Gain 1 Stamina or 1 Sanity?",
+          "1 Stamina",
+          "1 Sanity",
+          events.Gain(owner, {"stamina": 1}),
+          events.Gain(owner, {"sanity": 1}),
+      )
+    # While the text doesn't explicitly say to exhaust, if you don't,
```
fingoldfish, line 213: Thinking about this, maybe it would make sense to turn this into a trigger instead of a usable trigger (with the option to gain nothing). Although, unlike the Ruby, it can't be traded before someone else can use it (since trading only happens during movement). Maybe I'm overthinking it. If we find some way for players to trade during upkeep, the fact that it technically doesn't exhaust could be important.

eldritch/items/unique.py
```diff
@@ -56,3 +58,228 @@ def get_trigger(self, event, owner, state):
...
+        owner
+    )
+
+
+class GateBox(Item):
+  def __init__(self, idx):
+    super().__init__("Gate Box", idx, "unique", {}, {}, None, 4)
+
+  def get_interrupt(self, event, owner, state):
+    if not isinstance(event, events.GateChoice) or not event.character == owner:
```
fingoldfish, line 166: I would also make sure that this GateChoice is attached to a Return event. We currently don't use GateChoice anywhere else, but it's possible that we will (or create subclasses of GateChoice similar to NearestGateChoice).

eldritch/items/unique.py
```diff
@@ -56,3 +58,228 @@ def get_trigger(self, event, owner, state):
...
+      return None
+    if event.gate_name is None:
+      return None
+    return events.Sequence([
```
fingoldfish, line 170: I don't think this will do what you want. Events that reference the GateChoice don't have their references updated, so they will be referencing a cancelled GateChoice and do the wrong thing. Instead, you probably want to modify the GateChoice to set gate_name to None.

eldritch/items/unique.py
```diff
@@ -56,3 +58,228 @@ def get_trigger(self, event, owner, state):
...
+
+    if self.exhausted:
+      return None
+
+    print(owner.movement_points)
+    if owner.movement_points < movement_cost:
+      return None
+
+    rolls = events.DiceRoll(owner, 2)
+    two_success = events.Draw(owner, "spells", 2, keep_count=2)
```
fingoldfish, line 84: What do you think of the wording on this? "For every success, draw 1 spell." Draw them both together? Or is this two separate draws? I think I agree with your current implementation (drawn together).

eldritch/items/unique.py
```diff
@@ -1,9 +1,11 @@
-from eldritch import events
-from .base import Weapon, OneshotWeapon
+from eldritch import events, values
+from .base import Item, Weapon, OneshotWeapon
 
 __all__ = [
     "CreateUnique", "EnchantedKnife", "EnchantedBlade", "HolyWater", "MagicLamp", "MagicPowder",
-    "SwordOfGlory"
+    "SwordOfGlory",
```
fingoldfish, line 6: PallidMask?

eldritch/events.py
```diff
@@ -3231,6 +3239,10 @@ def resolve(self, state):
       assert self.combat.is_done()
       idx = self.monsters.index(self.combat.monster)
       self.done[idx] = True
+      for idx, monster in enumerate(self.monsters):
+        # Maybe an item defeated these monsters
```
fingoldfish, line 3243: If you use the flute to defeat the monster, but the monster is endless or a pinata, then this won't work the way you want it to. Maybe check that the place of the monster == the character's place instead.

eldritch/events.py
```diff
@@ -3247,8 +3259,9 @@ def resolve(self, state):
       annotations = ["Done" if self.done[idx] else None for idx, _ in present_monsters]
       prompt = "Choose a monster to fight or evade"
       none_choice = "Ignore All" if self.auto_evade else None
-      self.choice = MonsterChoice(self.character, prompt, shown_monsters, annotations, none_choice)
-      state.event_stack.append(self.choice)
+      if shown_monsters:
```
fingoldfish, line 3262: If shown_monsters has nothing, shouldn't all(self.done) be true? I guess I'm worried about a situation where we might show no monsters in the choice, but the event won't think it's "done" yet.

eldritch/items/unique.py
petercombs: I think it's just a single draw.  You don't want the Magician to be able to draw 2 extra cards.

eldritch/test_unique_items.py
```diff
@@ -0,0 +1,492 @@
...
+    super().setUp()
+    self.key = items.SilverKey(0)
+    self.char.possessions.append(self.key)
+    combat = events.Combat(
+        self.char, monsters.DreamFlier()
+    )
+    self.state.event_stack.append(combat)
+    evade_choice = self.resolve_to_choice(events.FightOrEvadeChoice)
+    evade_choice.resolve(self.state, "Flee")
+
```
fingoldfish, line 477: Can you also add a test that validates that you can use the silver key on the initial evade check that you make for an EvadeOrCombat? And also validate that if you use the silver key before entering combat that your movement does not end?

eldritch/items/unique.py
```diff
@@ -56,3 +70,257 @@ def get_trigger(self, event, owner, state):
...
+
+    seq = [
+        events.DiscardSpecific(owner, [self]),
+        events.PassCombatRound(combat_round)
+    ]
+    for monster in state.monsters:
+      if monster.place == owner.place and monster != event.monster:
+        seq.append(events.TakeTrophy(owner, monster))
+    seq.append(events.Loss(owner, {"stamina": 3, "sanity": 3}))
+    return events.Sequence(
```
fingoldfish, line 299: That's an interesting
multi line
statement.

eldritch/test_unique_items.py
```diff
@@ -0,0 +1,492 @@
...
+    fight_or_flee.resolve(self.state, "Fight")
+    fight_or_flee = self.resolve_to_choice(events.FightOrEvadeChoice)
+    fight_or_flee.resolve(self.state, "Fight")
+
+  def testNotEnoughSanity(self):
+    self.maniac.place = self.char.place
+    self.char.sanity = 2
+    self.enterCombat()
+    with self.assertRaises(AssertionError):
+      self.resolve_to_usable(0, "Flute0")
```
fingoldfish, line 312: Same comment as other tests.

eldritch/test_unique_items.py
```diff
@@ -0,0 +1,492 @@
...
+    self.cultist.place = self.char.place
+    self.char.stamina = 3
+    self.char.sanity = 3
+    self.enterCombat()
+    flute = self.resolve_to_usable(0, "Flute0")
+    self.state.event_stack.append(flute)
+    self.resolve_until_event_type(events.Devoured)
+    self.assertFalse(self.maniac.place)
+    self.assertFalse(self.cultist.place)
+
```
fingoldfish, line 332: And also one test where you go insane but keep the trophies?

eldritch/monsters.py
```diff
@@ -131,7 +131,15 @@ def movement(self, state):
         return movement
     return "normal"
 
-  def get_interrupt(self, event, state):  # pylint: disable=unused-argument
+  def get_interrupt(self, event, state):
+    if (isinstance(event, events.TakeTrophy)
```
fingoldfish, line 135: I am debating the merits of using an interrupt vs hard-coding this into TakeTrophy. If you hard-code this into TakeTrophy, you have to make a distinction between the behavior of TakeTrophy and ForceTakeTrophy. If you use an interrupt, you have to make sure that all monsters that could possibly be endless call return super().get_interrupt() when they don't have any interrupts. I'm not sure that I prefer one over the other, so I'm just making a mental note that I may want to look at this again in the future.

eldritch/monsters.py
```diff
@@ -131,7 +131,15 @@ def movement(self, state):
...
+            and self.has_attribute("endless", state, event.character)):
+      return events.Sequence(
+          [
+              events.CancelEvent(event),
+              events.ReturnMonsterFromBoard(event.character, self, ),
```
fingoldfish, line 140: Add a unit test for this?

fingoldfish (changes requested):

eldritch/events.py
```diff
@@ -2293,13 +2303,18 @@ def __init__(self, character, check_type, modifier, attributes=None):
     self.modifier = modifier
     self.attributes = attributes
     self.dice: Optional[Event] = None
+    self.pass_check: Optional[Event] = None
```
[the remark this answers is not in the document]
petercombs: PassCheck sets `Check.pass_check = self`.  I can't think of a reason why you'd  need access to the pass_check object from the check that's being passed, but that may just be my lack of imagination.

eldritch/events.py
```diff
@@ -2293,13 +2303,18 @@ def __init__(self, character, check_type, modifier, attributes=None):
...
     self.roll = None
     self.successes = None
     self.spend: Optional[SpendChoice] = None
     self.bonus_dice: Optional[Event] = None
     self.done = False
 
   def resolve(self, state):
+    if self.pass_check is not None:
+      self.successes = float("inf")
```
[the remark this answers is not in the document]
petercombs: I think the only thing that can PassCheck an event is the Blue Watcher, which can only pass combat checks, or fight and lore checks to close gates. However, the watcher _does_  need a check to make sure we can't use it in a combat check against an Ancient one.

eldritch/items/base.py
```diff
@@ -20,6 +20,9 @@ def __init__(
   def hands_used(self):
     return self.hands if self.active else 0
 
+  def get_max_token_event(self, token_type, owner):
```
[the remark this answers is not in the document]
petercombs: The way I figured it, tokens are used for: 

1. Multi-use items, which all Discard after a fixed number, 
2. A Water St encounter in Royal Harbor expansion, which returns to the box,
3. Personal stories, which are going to be different for every one, and
4. Some rumors (e.g. Mythos27), which are handled using a different mechanism

eldritch/items/unique.py
petercombs: I figured that `loss_type in ("sanity", "stamina")` doesn’t include clues.

eldritch/test_unique_items.py
```diff
@@ -0,0 +1,643 @@
...
+    loss = events.Loss(self.char, {"stamina": values.Calculation(die, "successes")})
+    self.state.event_stack.append(events.Sequence(
+        [die, loss], self.char
+    ))
+    with mock.patch.object(events.random, "randint", new=mock.MagicMock(return_value=3)):
+      self.resolve_until_done()
+    self.assertEqual(self.char.stamina, 5)
+    self.assertIn(self.statue, self.char.possessions)
+
+  def testJewelryAndStatue(self):
```
fingoldfish, line 581: Be sure to test both directions - if the obsidian statue is used, the jewelry should become unusable.

fingoldfish (changes requested):
Just a few more things.

eldritch/events.py
```diff
@@ -3144,6 +3164,33 @@ def annotations(self, state):
...
+    self.original_choice = original_choice
+    self.changes = changes
+    assert len(changes)
+    assert {"prompt", "gate_name", "none_choice", "annotation"}.issuperset(changes.keys())
+    self.done = False
+
+  def resolve(self, state):
+    for key, value in self.changes.items():
+      setattr(self.original_choice, key, value)
+    self.original_choice.compute_choices(state)
```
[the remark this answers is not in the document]
petercombs: Removing it causes `testTradeBeforeReturnMultipleGates` to fail at 
```
      self.assertEqual(nun_choice.character, nun)
>     self.assertSequenceEqual(nun_choice.choices, ["Diner", "Woods"])
E     AssertionError: Sequences differ: ['Woods'] != ['Diner', 'Woods']
E
E     First differing element 0:
E     'Woods'
E     'Diner'
E
E     Second sequence contains 1 additional elements.
E     First extra element 1:
E     'Woods'
```

fingoldfish (changes requested):
Requesting changes again - looks like there are 14 comments that got missed? In what I'm sure is a pure coincidence, they appear to be the same 14 comments that github auto-collapses in a section that says "14 hidden conversations" on the conversation page for this PR.

eldritch/events.py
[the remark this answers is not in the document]
fingoldfish: So I tried this out - it looks like there are two things going on here. The first is that after Dummy hands the GateBox to the Nun, he is still able to trade - you must set done_using[0] to True to continue the test normally.

The test currently expects the GateBox to be usable for the Nun. But this is only happening because the Dummy hasn't successfully returned. Once you fix the issue above and the Dummy successfully returns, the GateBox will be unconditional and the test will pass.

But this test has uncovered a second issue. You see, compute_choices for the GateChoice has a special side effect when there is only one choice - it will automatically choose the return destination. This was implemented based on feedback I received when we were last testing it out. So in this case, the GateChoice will compute choices, see that there is only one choice, and set event.choice = the only choice. Then, the GateBox comes along and changes event.gate_name to None. But the event loop sees the GateChoice as already being done, so it doesn't bother to call compute_choices again.

Stick a TODO in there for me to fix. It's a mess that I need to untangle.

eldritch/test_unique_items.py
```diff
@@ -0,0 +1,665 @@
...
+    self.enterCombat(0)
+    flute = self.resolve_to_usable(0, "Flute0")
+    self.state.event_stack.append(flute)
+    self.resolve_until_done()
+    self.assertEqual(self.char.stamina, 2)
+    self.assertEqual(self.char.sanity, 2)
+    self.assertIn(self.maniac, self.char.trophies)
+    self.assertNotIn(pinata, self.char.trophies)
+    self.assertNotIn(endless, self.char.trophies)
+    self.assertFalse(pinata.place)
```
fingoldfish, line 458: Nit: use assertIsNone when it should be none instead of just false-y

fingoldfish (approved):

PR #159, opened 2022-09-19
EventMonster to allow facing single-rounds of combat in Events

petercombs:
This actually fixes a bug with the Graveyard3 encounter that weapons weren't usable, while simplifying the Bank and Hospital encounters.

### DIFF
--- a/eldritch/events.py
+++ b/eldritch/events.py
@@ -3479,7 +3479,7 @@ class PassEvadeRound(Event):
 
 class CombatRound(Event):
 
-  def __init__(self, character, monster):
+  def __init__(self, character, monster, deactivate=False):
     super().__init__()
     self.character = character
     self.monster = monster
@@ -3492,6 +3492,7 @@ class CombatRound(Event):
     self.defeated = None
     self.damage: Optional[Event] = None
     self.pass_combat: Optional[Event] = None
+    self.deactivate = deactivate
     self.done = False
 
   def resolve(self, state):
@@ -3514,6 +3515,14 @@ class CombatRound(Event):
           self.character, "combat", self.monster.difficulty("combat", state, self.character), attrs)
     if not self.check.is_done():
       state.event_stack.append(self.check)
+      return
+
+    if self.deactivate and not isinstance(self.deactivate, Event):
+      self.deactivate = Sequence(
+          [DeactivateItems(self.character), DeactivateSpells(self.character)],
+          self.character
+      )
+      state.event_stack.append(self.deactivate)
       return
 
     if self.defeated is None:
--- a/eldritch/monsters.py
+++ b/eldritch/monsters.py
@@ -427,6 +427,29 @@ def Zombie():
   )
 
 
+class EventMonster(Monster):
+  """A pseudo-monster used for events like Bank3 and Hospital2"""
+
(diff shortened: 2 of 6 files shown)


eldritch/monsters.py
```diff
@@ -427,6 +427,29 @@ def Zombie():
   )
 
 
+class EventMonster(Monster):
+  """A pseudo-monster used for events like Bank3 and Hospital2"""
+
+  def __init__(self, name, rating, pass_event, fail_event, toughness=1, attributes=None):
+    super().__init__(
+        name, "normal", "moon", {"evade": 0} | rating,
```
fingoldfish, line 435: Using `|` on dictionaries is a feature introduced in python 3.9. To preserve 3.6 compatibility (actually, I should update everything to use 3.7 since 3.6 has been end of lifed), you can use something like `dict(evade=0, **rating)`.

eldritch/monsters.py
```diff
@@ -427,6 +427,29 @@ def Zombie():
   )
 
 
+class EventMonster(Monster):
+  """A pseudo-monster used for events like Bank3 and Hospital2"""
+
+  def __init__(self, name, rating, pass_event, fail_event, toughness=1, attributes=None):
```
fingoldfish, line 433: Would we ever want a toughness other than 1 or attributes for an event monster? I suppose it doesn't hurt to have them available.

eldritch/monsters.py
```diff
@@ -427,6 +427,29 @@ def Zombie():
...
+        {"combat": 0, }, toughness, attributes or set()
+    )
+    self.pass_event = pass_event
+    self.fail_event = fail_event
+
+  def get_trigger(self, event, state):
+    if isinstance(event, events.PassCombatRound):
+      return self.pass_event
+
+    if not isinstance(event, events.CombatRound) or not event.check.is_done():
```
fingoldfish, line 445: For safety, also please make sure that event.check is not None before doing the rest of these checks. Should avoid some unpleasant surprises.

eldritch/monsters.py
```diff
@@ -427,6 +427,29 @@ def Zombie():
...
+      return None
+
+    if (event.check.successes or 0) < self.toughness(state, event.character):
+      return events.Sequence([events.CancelEvent(event), self.fail_event], event.character)
```
fingoldfish, line 449: This cancellation seems superfluous - when you get here, `event` has finished (guaranteed since we are calling `get_trigger`), so you would be cancelling an event that is already done. I think this should work just fine without it.

eldritch/events.py
```diff
@@ -3516,6 +3517,14 @@ def resolve(self, state):
       state.event_stack.append(self.check)
       return
 
+    if self.deactivate and not isinstance(self.deactivate, Event):
```
fingoldfish, line 3520: You could probably create the deactivate event in the constructor, given that it doesn't need any information about what happened.
That being said, I think the game logic will automatically deactivate items at the end of a combat round. You might be able to only deactivate spells here? Or maybe you want to update the [spell deactivation](https://github.com/gummyboars/lunboks/blob/master/eldritch/eldritch.py#L669) to also work on CombatRounds against EventMonsters? Maybe that would be too much. Try out just deactivating the spells here and see if that works.

fingoldfish (approved):
LGTM

PR #171, opened 2022-11-14
First stab at Mythos 16-30

petercombs:


eldritch/mythos.py
```diff
@@ -412,6 +412,155 @@ def get_trigger(self, event, state):
...
+        {"square", "diamond"}, {"circle"},
+        "urban", activity_location="Uptown"
+    )
+
+  def get_trigger(self, event, state):
+    if not isinstance(event, events.Movement):
+      return None
+    if event.character.place.name != self.activity_location:
+      return None
+    return events.Purchase(event.character, "unique", draw_count=2, keep_count=2)
```
fingoldfish, line 428: Technically, the card says that they _may_ draw 2 unique items and purchase some. I think the only time this might come into play is if there is something on the bottom of the deck that the Drifter wanted to scrounge up?

eldritch/mythos.py
```diff
@@ -412,6 +412,155 @@ def get_trigger(self, event, state):
...
+    )
+
+  def get_trigger(self, event, state):
+    if not isinstance(event, events.Movement):
+      return None
+    if event.character.place.name != self.activity_location:
+      return None
+    check = events.Check(event.character, "lore", -1)
+    gain = events.Gain(event.character, {"clues": 1})
+    return events.PassFail(event.character, check, gain, events.Nothing)
```
fingoldfish, line 446: The last argument should be events.Nothing() instead of events.Nothing

eldritch/mythos.py
```diff
@@ -412,6 +412,155 @@ def get_trigger(self, event, state):
...
+    return None
+
+
+class Mythos20(Headline):
+  def __init__(self):
+    super().__init__("Mythos20", "Cave", "Roadhouse", {"moon"}, {"plus"})
+
+  def create_event(self, state):
+    seq = super().create_event(state)
+    seq.events.append(events.ReturnToCup(from_places=[
```
fingoldfish, line 492: ReturnToCup supports `from_places=["streets"]` (for streets and locations).

eldritch/mythos.py
```diff
@@ -412,6 +412,155 @@ def get_trigger(self, event, state):
...
+    return seq
+
+
+class Mythos24(Headline):
+  def __init__(self):
+    super().__init__("Mythos24", "WitchHouse", "Cave", {"circle"}, {"square", "diamond"})
+
+  def create_event(self, state):
+    seq = super().create_event(state)
+    seq.events.append(events.ReturnToCup(from_places=["Easttown", "Roadhouse", "Diner", "Police"]))
```
fingoldfish, line 537: You forgot to return seq here.

eldritch/mythos.py
```diff
@@ -412,6 +412,155 @@ def get_trigger(self, event, state):
...
+
+
+class Mythos25(Environment):
+  def __init__(self):
+    super().__init__("Mythos25", "Woods", "Square", {"square", "diamond"}, {"circle"}, "weather")
+
+  def get_modifier(self, thing, attribute):
```
fingoldfish, line 544: This should only affect investigators in the city, not in other worlds. See Mythos6 for an example.

eldritch/mythos.py
```diff
@@ -412,6 +412,155 @@ def get_trigger(self, event, state):
...
+    if attribute == "fight":
+      return -1
+    if attribute == "lore":
+      return 1
+    if thing.name == "Flame Matrix" and attribute == "toughness":
```
fingoldfish, line 549: Remember to use `getattr(thing, "name", None)`, since get_modifier may be called on things that don't have a name.

eldritch/mythos.py
```diff
@@ -456,6 +605,44 @@ def get_trigger(self, event, state):
     return super().get_trigger(event, state)
 
 
+class Mythos28(Headline):
+  def __init__(self):
+    super().__init__("Mythos28", "Graveyard", "Isle", {"hex"}, {"slash", "triangle", "star"})
+
+  def create_event(self, state):
+    first_player = state.characters[state.first_player]
```
fingoldfish, line 613: Remember to check whether the first player is devoured before cursing them.

eldritch/mythos.py
```diff
@@ -456,6 +605,44 @@ def get_trigger(self, event, state):
...
+    seq.events.extend([check, pass_fail])
+    return seq
+
+
+class Mythos29(Environment):
+  def __init__(self):
+    super().__init__("Mythos30", "Society", "Lodge", {"plus"}, {"moon"}, "weather")
+
+  def get_override(self, thing, attribute):
+    if isinstance(thing, monsters.Monster) and attribute == "fast":
```
fingoldfish, line 626: Can you add a test for this?

eldritch/mythos.py
```diff
@@ -456,6 +605,44 @@ def get_trigger(self, event, state):
...
+      return "normal"
+    return None
+
+  def get_interrupt(self, event, state):
+    if isinstance(event, events.Movement):
+      return events.ChangeMovementPoints(event.character, -1)
```
fingoldfish, line 632: Note that because this is an interrupt, it will run before Movement does. But Movement is the event that actually grants movement points. So I think you want to interrupt CityMovement, not Movement. Please add a test.

eldritch/test_mythos.py
```diff
@@ -2014,6 +2017,93 @@ def testNormal(self):
...
+  def drawMythos(self):
+    self.state.mythos.append(self.mythos)
+    self.state.event_stack.append(Mythos(self.char))
+    self.resolve_until_done()
+
+  def testNoMonstersReturned(self):
+    self.drawMythos()
+    self.assertEqual(self.state.terror, 0)
+
+  def testOneMonsterReturned(self):
```
fingoldfish, line 2095: Add one test for the outskirts - nothing should be returned; the terror level should not go up.

fingoldfish (changes requested):
More tests wouldn't hurt.

eldritch/test_mythos.py
```diff
@@ -2055,6 +2058,130 @@ def testNormal(self):
...
+    choice = self.resolve_to_choice(CardSpendChoice)
+    self.assertEqual(choice.choices, ["Downtown Card", "Restore 1 Sanity", "Restore All Sanity"])
+    self.spend("dollars", 2, choice)
+    choice = self.resolve_to_choice(CardSpendChoice)
+    choice.resolve(self.state, "Restore All Sanity")
+    self.resolve_until_done()
+    self.assertEqual(self.char.sanity, 5)
+
+  def testCantGainFromOthers(self):
+    self.char.possessions.append(items.HealingStone(0))
```
fingoldfish, line 2106: TODO for me: maybe the healing stone shouldn't even let you try to gain sanity if this card is in play.

eldritch/mythos.py
```diff
@@ -412,6 +422,163 @@ def get_trigger(self, event, state):
...
+        isinstance(event, events.GainOrLoss)
+        and values.Calculation(event.gains["sanity"], operand=values.operator.gt, right=0).value(
+            state)
+        and (event.source is None or event.source.name not in ("Psychology", "Asylum"))
+    ):
+      return events.GainPrevention(self, event, "sanity", event.gains["sanity"])
+    return None
+
+
+class Mythos19(Headline):
```
fingoldfish, line 479: Can you also add tests for this? Maybe stick a monster and a character in the merchant streets. Validate that this card gets removed at the end of the next mythos?

fingoldfish (changes requested):
Nitpicks.

eldritch/events.py
```diff
@@ -901,10 +904,20 @@ def is_resolved(self):
   def log(self, state):
     if self.prevented is None:
       amount = self.amount.value(state) if isinstance(self.amount, values.Value) else self.amount
-      return f"{self.prevention_source.name} will prevent {amount} {self.attribute} loss"
+      return (f"{self.prevention_source.name} will prevent {amount} {self.attribute}"
```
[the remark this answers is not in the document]
petercombs: I was aiming for "will prevent X stamina from losses", which is only slightly better.

eldritch/events.py
```diff
@@ -1301,6 +1313,17 @@ def __init__(self, character, location_name):
     self.done = False
 
   def resolve(self, state):
+    if (isinstance(self.location_name, str)
+            and getattr(state.places[self.location_name], "closed_until", None) is not None):
+      destinations = {place.name for place in state.places[
```
fingoldfish, line 1318: Shouldn't you also check for the other places being closed and only offer the player open locations?

fingoldfish (changes requested):
Just a few things remaining.

eldritch/events.py
```diff
@@ -1301,6 +1313,17 @@ def __init__(self, character, location_name):
     self.done = False
 
   def resolve(self, state):
+    if (isinstance(self.location_name, str)
+            and getattr(state.places[self.location_name], "closed_until", None) is not None):
```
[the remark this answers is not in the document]
petercombs: Because I forgot that was an option!

eldritch/events.py
petercombs: I've now added a test to ensure they can't choose a closed location.

fingoldfish (approved):

PR #183, opened 2023-01-21
If an encounter hasn't been fully implemented, draw another one

petercombs:
I also implemented a few of the outstanding todos that required little thought.

### DIFF
--- a/eldritch/encounters.py
+++ b/eldritch/encounters.py
@@ -22,7 +22,7 @@ class EncounterCard:
   def encounter_event(self, character, location_name):
     if location_name not in self.encounters:
       print(f"TODO: missing encounters for {self.name} at {location_name}")
-      return events.Nothing()
+      return events.Unimplemented()
     return self.encounters[location_name](character)
 
 
@@ -119,11 +119,11 @@ def Roadhouse3(char):
 
 
 def Roadhouse4(char):
-  return events.Nothing()  # TODO buying stuff
+  return events.Purchase(char, "common", 3, keep_count=3, discount_type="fixed", discount=-1)
 
 
 def Roadhouse5(char):
-  return events.Nothing()  # TODO monster cup
+  return events.MonsterAppears(char)
 
 
 def Roadhouse6(char):
@@ -299,7 +299,7 @@ def Sanctum5(char):
 
 
 def Sanctum6(char):
-  return events.Nothing()  # TODO: A monster appears
+  return events.MonsterAppears(char)
 
 
 def Sanctum7(char):
@@ -312,7 +312,8 @@ def Sanctum7(char):
       char, "Participate in a gating ceremony?", ["Yes", "No"], spends=[spend, None],
   )
   cond = events.Conditional(char, choice, "choice_index", {0: ceremony, 1: events.Nothing()})
-  return events.Sequence([choice, cond], char)
+  # return events.Sequence([choice, cond], char)
+  return events.Unimplemented()
 
 
 def WitchHouse1(char):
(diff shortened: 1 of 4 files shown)


eldritch/events.py
```diff
@@ -1718,6 +1722,12 @@ def resolve(self, state):
 
     encounters = [
         card.encounter_event(self.character, location_name) for card in self.draw.cards]
+    if any(isinstance(enc, Unimplemented) for enc in encounters):
```
fingoldfish, line 1725: This should go above the section where we short-circuit if there is only one card drawn (the line that says TODO: test this). The short circuit only happens in test mode, so your two options are to turn test mode off to test this functionality or to move the if condition up.

eldritch/events.py
```diff
@@ -1806,6 +1816,10 @@ def resolve(self, state):
       return
 
     encounters = [card.encounter_event(self.character, world_name) for card in self.cards]
+    if any(isinstance(enc, Unimplemented) for enc in encounters):
+      self.draw = None
```
fingoldfish, line 1820: Does this work? Because when you get here, len(self.cards) == self.draw_count, so you won't draw another card the next time resolve() is called. And you haven't removed anything from cards, so the player can still select an unimplemented encounter. I think you would need to put the cards back in the deck (maybe only put the unimplemented cards back?) before doing this.

eldritch/events.py
```diff
@@ -1806,6 +1816,10 @@ def resolve(self, state):
       return
 
     encounters = [card.encounter_event(self.character, world_name) for card in self.cards]
+    if any(isinstance(enc, Unimplemented) for enc in encounters):
```
fingoldfish, line 1819: Same comment about putting this before the short-circuit.

eldritch/encounters.py
```diff
@@ -119,11 +119,11 @@ def Roadhouse3(char):
 
 
 def Roadhouse4(char):
-  return events.Nothing()  # TODO buying stuff
+  return events.Purchase(char, "common", 3, keep_count=3, discount_type="fixed", discount=-1)
```
fingoldfish, line 122: Please add tests. For the encounters where a monster appears, nothing fancy is needed - maybe just check that the player gets to a fight or evade choice?

eldritch/encounters.py
```diff
@@ -372,7 +373,7 @@ def Cave1(char):
   monster = events.Sequence(
       [
           san_loss,
-          # TODO: Implement a monster appears
+          events.MonsterAppears(char),
```
fingoldfish, line 376: Now that we have this, can we change the way this has been indented?

fingoldfish (changes requested):

eldritch/events.py
petercombs: Given that "ignore cards that aren't implemented" isn't a proper feature of original rules, I think removing all cards (not just the un implemented ones) is fine, and saves having to muck about with the middles of lists.

fingoldfish (approved):

eldritch/encounters.py
```diff
@@ -22,7 +22,7 @@ def add_encounter(self, location_name, encounter_creator):
   def encounter_event(self, character, location_name):
     if location_name not in self.encounters:
       print(f"TODO: missing encounters for {self.name} at {location_name}")
-      return events.Nothing()
+      return events.Unimplemented()
```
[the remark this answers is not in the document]
petercombs: Only if none of the cards for a given location had the encounter implemented. Otherwise, `Encounter.resolve()` will just keep drawing cards until it finds one that works.

PR #184, opened 2023-01-24
Gate Encounters 1-15

petercombs:


eldritch/events.py
```diff
@@ -2718,8 +2722,10 @@ def log(self, state):
     return ""
 
 
-def PassFail(character, condition, pass_result, fail_result):
-  outcome = Conditional(character, condition, "successes", {0: fail_result, 1: pass_result})
+def PassFail(character, condition, pass_result: Event, fail_result: Event, min_successes=1):
```
petercombs, line 2725: @fingoldfish Thoughts on putting the check difficulty here versus in the check itself somehow? Could always have the PassFail function as something like 
```python
outcome = Conditional(character, condition, "successes", {0: fail_result, condition.difficulty: pass_result})
```

What else should get a difficulty attribute if we go that route?

eldritch/events.py
```diff
@@ -93,9 +93,18 @@ def log(self, state) -> str:
...
+    self._choices = None
+
+  @property
+  def choices(self):
+    return self._choices
 
   @abc.abstractmethod
-  def resolve(self, state, choice=None) -> NoReturn:  # pylint: disable=arguments-differ
+  def resolve(
+      self, state, choice=None
```
petercombs, line 106: I split this line so that I could give the type hint that state is an eldritch.GameState, but then ended up in a circular import purgatory...

eldritch/events.py
```diff
@@ -125,6 +134,10 @@ def log(self, state):
     return ""
 
 
+class Unimplemented(Nothing):
```
petercombs, line 137: Borrowed from the other branch, without the "Skip unimplemented cards" logic.

eldritch/events.py
```diff
@@ -93,9 +93,18 @@ def log(self, state) -> str:
 
 
 class ChoiceEvent(Event):
+  def __init__(self):
+    super().__init__()
+    self._choices = None
+
+  @property
+  def choices(self):
```
petercombs, line 101: Every choice event that we have has `choices`, unsurprisingly.  But adding it into the base class shuts my IDE up when it warns me about using a ChoiceEvent's choices.

eldritch/test_events.py
```diff
@@ -88,7 +88,7 @@ def resolve_until_done(self):
       not_finished["stack"] = self.state.event_stack
     self.assertFalse(not_finished)
 
-  def resolve_to_choice(self, event_class):
+  def resolve_to_choice(self, event_class) -> events.ChoiceEvent:
```
petercombs, line 91: This mainly helps my IDE warn me when I try to do `choice.resolve("thing")` instead of `choice.resolve(self.state, "thing")`

eldritch/values.py
```diff
@@ -1,12 +1,17 @@
...
 import collections
+import math
 import operator
 
 from eldritch import places
 
 OPER_MAP = {"at least": operator.ge, "at most": operator.le, "exactly": operator.eq}
 
 
+def ceildiv(lhs, rhs):
```
petercombs, line 11: Seems simper to have a single operator rather than a chain of Calculations for rounding up.

eldritch/gate_encounters.py
```diff
@@ -20,64 +23,393 @@ def encounter_event(self, character, world_name):
...
+  # check = events.Check(char, "speed", -2)
+  # return events.Sequence(
+  #     [
+  #       events.PassFail(
+  #           char,
+  #           check,
+  #           events.Return(char, char.place.info.name),
+  #           events.LostInTimeAndSpace(char),
+  #       ),
+  #       events.CloseGate(char, "the one you entered"),
```
petercombs, line 65: I think this is the real showstopper for this event... Or does it really just mean close the gate you returned through? Because it's possible that the gate you entered the other world through is already closed, right?

eldritch/gate_encounters.py
```diff
@@ -20,64 +23,393 @@ def encounter_event(self, character, world_name):
...
+  weapon = events.Draw(char, "common", 1, target_type=items.Weapon)
+  return events.PassFail(char, check, weapon, events.Nothing())
+
+
+def Pluto12(char) -> events.Event:
+  check = events.Check(char, "will", -2)
+  san_loss = events.Loss(char, {"sanity": 2})
+  spell_loss = events.LoseItems(char, 2, "Lose spells", {"spells"})
+  prereq = values.ItemDeckPrerequisite(char, "spells", threshold=2)
+  choice = events.BinaryChoice(
```
fingoldfish, line 255: Instead of creating a second pass/fail, you can pass the prereq object directly into BinaryChoice (it will get attached to the ChoiceEvent, and the choice will be greyed out if the user does not meet the prerequisite)

eldritch/gate_encounters.py
```diff
@@ -20,64 +23,393 @@ def encounter_event(self, character, world_name):
...
+def GreatHall12(char) -> events.Event:
+  return events.MonsterAppears(char)
+
+
+def Other12(char) -> events.Event:
+  check = events.Check(char, "sneak", -2)
+  return events.PassFail(char, check, events.Nothing(), events.Delayed(char))
+
+
+def City13(char) -> events.Event:
```
fingoldfish, line 276: Wow, this encounter is terrible.

eldritch/gate_encounters.py
```diff
@@ -20,64 +23,393 @@ def encounter_event(self, character, world_name):
...
+      char, "Trade with the dangerous hooved folk?",
+      "Yes", "No",
+      events.PassFail(char, check, trade, lost), events.Nothing(),
+  )
+  return choice
+
+
+def Other13(char) -> events.Event:
+  check = events.Check(char, "luck", -1)
+  return events.PassFail(char, check, events.Gain(char, {"clues": 1}), events.Nothing())
```
fingoldfish, line 300: The card says 2 clue tokens

eldritch/gate_encounters.py
```diff
@@ -20,64 +23,393 @@ def encounter_event(self, character, world_name):
...
+  rope = events.Return(char, char.place.info.name)
+  fall = events.Sequence([
+      events.Loss(char, {"stamina": 2}),
+      events.Delayed(char),
+  ], char)
+  return events.PassFail(char, check, rope, fall)
+
+
+def City15(char) -> events.Event:
+  check = events.Check(char, "lore", -2)
```
fingoldfish, line 330: This is actually a lore -2 [2] check

eldritch/gate_encounters.py
```diff
@@ -20,64 +23,393 @@ def encounter_event(self, character, world_name):
...
+  spells = events.Draw(char, "spells", 2)
+  return events.PassFail(char, check, spells, events.Nothing())
+
+
+def GreatHall15(char) -> events.Event:
+  check = events.Check(char, "lore", 2)
+  return events.Sequence([
+      check,
```
fingoldfish, line 338: Same comment as above - make this a conditional.

eldritch/test_gates.py
```diff
@@ -5,16 +5,19 @@
 import sys
 import unittest
 from unittest import mock
+from typing import cast
 
 # Hack to allow the test to be run directly instead of invoking python from the base dir.
+import eldritch.places
```
fingoldfish, line 11: Why did this move? And also get imported as eldritch.places instead of just places? This should probably get imported below the hack that changes the import paths to make this discoverable.

fingoldfish (changes requested):
First pass.

eldritch/gate_encounters.py
```diff
@@ -20,64 +23,393 @@ def encounter_event(self, character, world_name):
...
+
+
+def Other8(char) -> events.Event:
+  return events.Draw(char, "common", 1)
+
+
+def Dreamlands9(char) -> events.Event:
+  check = events.Check(char, "sneak", 0)
+  lose_item_count = values.Calculation(
+      values.ItemDeckCount(char, {"common", "unique", "spells", "tradables"}), None,
```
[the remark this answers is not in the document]
petercombs: I guess I interpreted it as `lose {half your items (your choice) and half your money}(rounded up)` rather than `lose {half your items (your choice)} and {half your money (rounded up)}`. Why didn't they write these in Lojban?

eldritch/events.py
petercombs: Leaving this link for my future reference. https://docs.python.org/3/library/typing.html#constant

eldritch/events.py
petercombs: I think I tried an attribute initially, but while most classes use it as an attribute, but CityMovement already has a property. I also don't remember exactly what the error was, but yeah, unintuitive.

eldritch/test_gates.py
```diff
@@ -83,6 +85,210 @@ def setUp(self):
...
+    with mock.patch.object(events.random, "randint", new=mock.MagicMock(side_effect=side_effect)):
+      self.resolve_until_done()
+    self.assertListEqual(self.char.possessions, [key])
+
+  def testThreeSuccessesFail(self):
+    side_effect = [5, 5, 5, 3] + [3, 3]
+    self.assertEqual(len(self.char.possessions), 0)
+    with mock.patch.object(events.random, "randint", new=mock.MagicMock(side_effect=side_effect)):
+      self.resolve_until_done()
+    self.assertEqual(len(self.char.possessions), 0)
```
fingoldfish, line 166: Maybe also assert on the location here?

fingoldfish (approved):

PR #187, opened 2023-02-02
Guardian Angel ability

petercombs:


eldritch/abilities.py
```diff
@@ -96,6 +96,26 @@ def __init__(self):
     super().__init__("Flux Stabilizer")
 
 
+class GuardianAngel(assets.Asset):
```
fingoldfish, line 99: Don't forget to add this to CreateSpecials at the bottom.

fingoldfish (changes requested):
LGTM as soon as tests and lint pass.

PR #188, opened 2023-03-25
Make Blessing/Curse an asset

petercombs:
Once I get the hang of this one, I can do loans, retainers, etc in one more PR.

### DIFF
--- a/eldritch/assets.py
+++ b/eldritch/assets.py
@@ -73,6 +73,9 @@ class Asset(metaclass=abc.ABCMeta):
   def get_max_token_event(self, token_type, owner):
     return events.Nothing()
 
+  def get_zero_tokens_event(self, token_type, owner):
+    return events.Nothing()
+
   def json_repr(self):
     return {attr: getattr(self, attr, None) for attr in self.JSON_ATTRS}
 
@@ -273,34 +276,47 @@ class Deputy(Card):
     return None
 
 
-class BlessingOrCurse(Card):
-  def __init__(self, kind, idx):
-    assert kind in ["Blessing", "Curse"]
-    super().__init__(kind, idx, "specials", {}, {})
-    self.opposite = "Curse" if kind == "Blessing" else "Blessing"
-    self.must_roll = False
+class SelfDiscardingCard(Card):
+  def __init__(self, name, idx, active_bonuses=None, passive_bonuses=None):
+    active_bonuses = active_bonuses or {}
+    passive_bonuses = passive_bonuses or {}
+    super().__init__(
+        name, idx, "specials", active_bonuses=active_bonuses, passive_bonuses=passive_bonuses
+    )
+    self.tokens["elder_sign"] = 0
 
   def get_trigger(self, event, owner, state):
     if isinstance(event, events.UpkeepActions) and event.character == owner:
-      if self.must_roll:
-        roll = events.DiceRoll(owner, 1)
-        return events.Conditional(
-            owner, roll, "sum",
-            {1: events.DiscardSpecific(owner, self), 2: events.Nothing()}
-        )
-      self.must_roll = True
+      if self.tokens["elder_sign"]:
+        return events.RollToLose(owner, self)
+      return events.AddToken(self, "elder_sign", owner)
+    return None
 
+
+class BlessingOrCurse(SelfDiscardingCard):
+  def __init__(self, name, idx):
+    assert name in ["Blessing", "Curse"]
+    super().__init__(name, idx)
+    self.opposite = "Curse" if name == "Blessing" else "Blessing"
+
+  def get_trigger(self, event, owner, state):
     if isinstance(event, events.KeepDrawn) and self.name in event.kept:
       selves = [p for p in event.character.possessions if p.name == self.name]
-      _, *duplicates = sorted(selves, key=lambda x: x.must_roll)
+      kept, *duplicates = sorted(selves, key=lambda x: x.tokens["elder_sign"])
       if duplicates:
-        return events.DiscardSpecific(event.character, duplicates)
+        return events.Sequence([
+            events.DiscardSpecific(event.character, duplicates),
+            events.RemoveToken(kept, "elder_sign", owner),
+        ], owner)
       if self.opposite in [p.name for p in event.character.possessions]:
         return events.Sequence([
-            events.DiscardNamed(event.character, self.name),
-            events.DiscardNamed(event.character, self.opposite),
+            events.LoseNamed(event.character, self.name),
+            events.LoseNamed(event.character, self.opposite),
         ], event.character)
-    return None
+    return super().get_trigger(event, owner, state)
+
+  def __repr__(self):
+    return f"<{self.name} - {self.tokens}>"
 
 
 def Blessing(idx):
--- a/eldritch/assets.py
+++ b/eldritch/assets.py
@@ -283,13 +283,13 @@ class SelfDiscardingCard(Card):
     super().__init__(
         name, idx, "specials", active_bonuses=active_bonuses, passive_bonuses=passive_bonuses
     )
-    self.tokens["elder_sign"] = 0
+    self.tokens["must_roll"] = 0
 
   def get_trigger(self, event, owner, state):
     if isinstance(event, events.UpkeepActions) and event.character == owner:
-      if self.tokens["elder_sign"]:
+      if self.tokens["must_roll"]:
         return events.RollToLose(owner, self)
-      return events.AddToken(self, "elder_sign", owner)
+      return events.AddToken(self, "must_roll", owner)
     return None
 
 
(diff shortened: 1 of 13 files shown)


eldritch/test_events.py
```diff
@@ -1286,32 +1286,34 @@ def testBecomeMember(self):
...
     self.assertEqual(self.char.bless_curse, 1)
-    self.assertEqual(self.char.bless_curse_start, self.state.turn_number + 2)
-    self.assertEqual(bless.change, 0)
+    self.assertEqual([p.name for p in self.char.possessions], ["Blessing"])
+    self.assertFalse(self.char.possessions[0].must_roll)
 
   def testCursedWhileBlessed(self):
+    self.state.event_stack.append(Bless(self.char))
+    self.resolve_until_done()
+    self.char.possessions[0].must_roll = True
```
fingoldfish, line 1306: Do you need the must_roll here?

eldritch/assets.py
```diff
@@ -273,6 +273,44 @@ def get_trigger(self, event, owner, state):
...
+
+  def get_trigger(self, event, owner, state):
+    if isinstance(event, events.UpkeepActions) and event.character == owner:
+      if self.must_roll:
+        roll = events.DiceRoll(owner, 1)
+        return events.Conditional(
+            owner, roll, "sum",
+            {1: events.DiscardSpecific(owner, self), 2: events.Nothing()}
+        )
+      self.must_roll = True
```
fingoldfish, line 291: I have a bad feeling that get_trigger may be called more than once. Can you add some tests that verify that the don't-roll-the-first-turn part is working?
You could also implement this as a token, and just return an AddToken event if the current count is 0 (and the roll to discard otherwise).

eldritch/assets.py
```diff
@@ -273,6 +273,44 @@ def get_trigger(self, event, owner, state):
...
+
+    if isinstance(event, events.KeepDrawn) and self.name in event.kept:
+      selves = [p for p in event.character.possessions if p.name == self.name]
+      _, *duplicates = sorted(selves, key=lambda x: x.must_roll)
+      if duplicates:
+        return events.DiscardSpecific(event.character, duplicates)
```
fingoldfish, line 297: This may be problematic. There are a few cards that say "if you lose your blessing, ...". Might make more sense to use an interrupt to change the KeepDrawn event? There's also the Nun's personal story, which adds a token each time a player is blessed. Do we think that story should advance if a player is blessed while they already have a blessing?

If no, then you might want to make this an interrupt - interrupt a KeepDrawn and then replace it with a different event (either change tokens, or discard specific). If yes, then we may need to make bless/curse its own event type to deal with this (or hard-code it into KeepDrawn, but I think that will get messy).

eldritch/assets.py
```diff
@@ -273,6 +273,44 @@ def get_trigger(self, event, owner, state):
...
+class BlessingOrCurse(Card):
+  def __init__(self, kind, idx):
+    assert kind in ["Blessing", "Curse"]
+    super().__init__(kind, idx, "specials", {}, {})
+    self.opposite = "Curse" if kind == "Blessing" else "Blessing"
+    self.must_roll = False
+
+  def get_trigger(self, event, owner, state):
+    if isinstance(event, events.UpkeepActions) and event.character == owner:
+      if self.must_roll:
```
fingoldfish, line 285: The other thing to look out for here is that must_roll does not reset if you discard a blessing/curse. So you might have to roll if you draw a blessing that someone else had already used and discarded.

fingoldfish (changes requested):

eldritch/assets.py
```diff
@@ -273,6 +273,44 @@ def get_trigger(self, event, owner, state):
     return None
 
 
+class BlessingOrCurse(Card):
```
fingoldfish, line 276: Oh, can you also add this to names.js - probably under otherNames.

eldritch/assets.py
petercombs: I would think that re-blessing a player would advance the nun's story, but should not count as them losing their blessing. From the FAQ:
> Treat the card as if it were newly acquired, so do not roll during the next upkeep to see if you lose it.

One way to implement this could be to make an empty subclass of DiscardSpecific LoseBlessingOrCurse, and then have the logic be:

- LoseBlessingOrCurse if you roll a 1
- LoseBlessingOrCurse if you draw the opposite
- DiscardSpecific(event.character, duplicates)

That way, cards that care that you're losing your blessing can look for the LoseBlessingOrCurse, but the behavior is ultimately the same.

eldritch/assets.py
```diff
@@ -273,6 +276,57 @@ def get_trigger(self, event, owner, state):
...
+  def get_trigger(self, event, owner, state):
+    if isinstance(event, events.KeepDrawn) and self.name in event.kept:
+      selves = [p for p in event.character.possessions if p.name == self.name]
+      kept, *duplicates = sorted(selves, key=lambda x: x.tokens["elder_sign"])
+      if duplicates:
+        return events.Sequence([
+            events.DiscardSpecific(event.character, duplicates),
+            events.RemoveToken(kept, "elder_sign", owner),
+        ], owner)
+      if self.opposite in [p.name for p in event.character.possessions]:
```
fingoldfish, line 311: How will this interact with the nun's personal story? If she gains a curse card, she fails. But in this case, she would gain a curse card, and then immediately discard it.

The more I look at the different possibilities here, the more I think we may actually need to hard-code this interaction (into KeepDrawn, maybe?). I can't think of an elegant way to do it.

eldritch/events.py
```diff
@@ -4472,6 +4508,51 @@ def animated(self):
     return True
 
 
+class RemoveToken(Event):
```
fingoldfish, line 4511: Would it be easier to just also account for zero tokens in the AddToken event, and allow AddToken to accept negative numbers (which it currently can, but we don't floor tokens at zero right now).

fingoldfish (changes requested):

eldritch/events.py
```diff
@@ -2374,6 +2358,55 @@ def animated(self):
...
+    super().__init__()
+    self.character = character
+    self.item = item
+    self.roll = None
+    self.lose = None
+    self.done = False
+
+  def resolve(self, state) -> NoReturn:
+    if self.roll is None:
+      self.roll = DiceRoll(self.character, 1)
```
fingoldfish, line 2384: Addendum: if you rebase, you will see some new params on DiceRoll. The args are `name`, which should be set to the name of the item you are rolling for, and `bad`, which should be set to a list of dice rolls that would result in a bad outcome (in this case, `[1]`).

These are UI-only parameters; they should have no effect on the game logic. The `name` param is also available on checks, but it can be omitted if the check/dice roll is the result of an encounter. The `bad` param is needed for any dice roll that is not checking success/failure, though in some cases it may be set to an empty list (i.e. roll a die and lose that much stamina).

eldritch/events.py
```diff
@@ -2361,6 +2354,55 @@ def animated(self):
     return True
 
 
+class LoseSpecific(DiscardSpecific):
```
fingoldfish, line 2357: I might do these the other way around, where DiscardSpecific is a subclass of LoseSpecific. This way, things that look for isinstance(event, DiscardSpecific) will not pick up on LoseSpecific, which looks like it is being used to discard blessings/curses without triggering any other events.

eldritch/events.py
```diff
@@ -1199,45 +1199,27 @@ def __init__(self, character):
...
-
-  def resolve(self, state):
-    old_val = self.character.bless_curse
-    new_val = min(max(old_val + self.adjustment, -1), 1)
-    self.character.bless_curse = new_val
-    self.change = new_val - old_val
-    if new_val != 0:
-      self.character.bless_curse_start = state.turn_number + 2
+    if positive:
+      self.card = "Blessing"
```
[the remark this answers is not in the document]
petercombs: I think my original vision was that if you were checking whether an event was a Bless or Curse, you could check for the event.card, but then I just made them their own class, because `if isinstance(event, Bless)` is nicer than `if isinstance(event, BlessCurse) and event.card == "Blessing"`. 

Removing the attribute.

eldritch/events.py
```diff
@@ -2361,6 +2354,55 @@ def animated(self):
...
+  items = values.NamedPossessions(character, name)
+  return LoseSpecific(character, items, to_box)
+
+
+class RollToLose(Event):
+  def __init__(self, character, item):
+    super().__init__()
+    self.character = character
+    self.item = item
+    self.roll = None
```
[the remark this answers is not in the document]
petercombs: So we would have something like: 
```python
def __init__(...):
  ...
  self.roll = None
  ...

def resolve(...):
  ...
  if not self.roll.is_resolved():
      state.event_stack.append(self.roll)
      return
```

I feel like the usual pattern in the other events is to check whether the children are None, and if so create the event.

eldritch/events.py
petercombs: As discussed, there is no difference between Losing and Discarding, so I'm getting rid of the class.

fingoldfish (approved):

eldritch/events.py
```diff
@@ -1199,45 +1199,27 @@ def __init__(self, character):
     super().__init__([ForceMovement(character, "Lost"), LoseTurn(character)], character)
 
 
-class BlessCurse(Event):
-
+class BlessCurse(Sequence):
```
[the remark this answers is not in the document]
petercombs: I wonder if you actually want to move some of the discard logic into SelfDiscardingCard class, to ensure you only have one loan/retainer? Maybe make a SingletonSelfDiscarding card subclass that those ones inherit from? Because the Blessing/Curse logic isn't the same. 

To be investigated another day.

PR #197, opened 2023-04-02
DEMO: Drifter and Nun stories

petercombs:


eldritch/test_events.py
```diff
@@ -149,8 +156,11 @@ def advance_turn(self, target_turn, target_phase):
           break
         continue
       if self.state.turn_phase == "upkeep":
-        self.assertIsInstance(self.state.event_stack[-1], events.SliderInput)
-        self.state.event_stack[-1].resolve(self.state, "done", None)
+        # self.assertIsInstance(self.state.event_stack[-1], events.SliderInput)
```
petercombs, line 159: Having items that are usable during upkeep breaks the `assertIsInstance(event, SliderInput)`. Maybe I should have an optional `may_break_early=False` arg that, if False will skip past SliderInputs, but not other Usables/chocies.

eldritch/test_events.py
```diff
@@ -92,6 +92,13 @@ def resolve_until_done(self):
       not_finished["stack"] = self.state.event_stack
     self.assertFalse(not_finished)
 
+  def resolve_to_slider_input(self) -> SliderInput:
+    # SliderInput has a different resolve() signature than a Choice
```
petercombs, line 96: Having a separate function makes my type checker happier.

eldritch/stories.py
```diff
@@ -0,0 +1,202 @@
...
+    if isinstance(event, events.Curse) and event.character == owner:
+      return events.CancelEvent(event)
+    return super().get_interrupt(event, owner, state)
+
+  def get_usable_trigger(self, event, owner, state):
+    if isinstance(event, events.DiceRoll) and state.turn_phase == "upkeep":
+      # Note: I see nothing that says it has to be one of her dice
+      # TODO: Allow user to select a die to reroll
+      to_reroll = [0]
+      # to_reroll = events.ChooseDie(event)
```
fingoldfish, line 150: I would look at the Bullwhip for an example of choosing a die to reroll.

eldritch/stories.py
```diff
@@ -0,0 +1,202 @@
...
+      return events.Sequence([
+          events.ExhaustAsset(owner, self),
+          events.RerollSpecific(state.event_stack[-2].character, state.event_stack[-2], to_reroll),
+      ], owner)
```
fingoldfish, line 154: Oof, RerollSpecific currently only works for checks; you may have to update it to work for either checks or DiceRolls.

eldritch/test_events.py
fingoldfish: Oh, I see. The story becomes usable because you roll the die. You could just put something like `if not isinstance(SliderInput): self.state.done_using[char_idx] = True; continue` in here.

eldritch/events.py
```diff
@@ -4142,6 +4142,50 @@ def log(self, state):
...
+      if self.gate.is_cancelled():
+        self.cancelled = True
+        return
+      self.gate = state.places[self.gate.choice].gate
+    if self.gate == "draw":
+      if state.gates:
+        self.gate = state.gates.popleft()
+      else:
+        # According to my reading, No Gate Markers only awakens the ancient one when a gate opens
+        self.gate = None
```
fingoldfish, line 4163: Maybe in this case, mark the event as cancelled and return early.

eldritch/events.py
```diff
@@ -4342,19 +4388,26 @@ def resolve(self, state: "eldritch.GameState"):
       if self.gate is None:
         self.cancelled = True
         return
+      state.places[self.location_name].gate = None
 
       if self.can_take:
-        # TODO: event for taking a gate trophy
-        self.character.trophies.append(self.gate)
+        # self.character.trophies.append(self.gate)
```
fingoldfish, line 4394: Remove comment?

eldritch/stories.py
```diff
@@ -0,0 +1,202 @@
...
+    super().__init__("Crime Doesn't Pay")
+
+  def get_in_play_event(self, owner: characters.Character):
+    return events.Loss(owner, {"dollars": owner.dollars})
+
+  def get_interrupt(self, event, owner, state):
+    if (
+        isinstance(event, events.GainOrLoss)
+        and event.character == owner
+        and "dollars" in event.gains
```
fingoldfish, line 113: Do we think this also affects trading money?

eldritch/test_stories.py
```diff
@@ -0,0 +1,195 @@
...
+    self.assertFalse(self.char.possessions[0].tokens["must_roll"])
+
+    with mock_randint(1) as mock_rand:
+      self.advance_turn(2, "movement")
+      self.assertIn(0, self.state.usables)
+      self.assertEqual(self.state.turn_phase, "upkeep")
+      self.assertEqual(mock_rand.call_count, 1)
+      self.assertIsInstance(self.state.event_stack[-1], events.DiceRoll)
+      self.assertEqual(self.state.event_stack[-1].roll, [1])
+      self.state.done_using[0] = True
```
fingoldfish, line 182: Can you add one test for rerolling the die?

fingoldfish (changes requested):

eldritch/stories.py
petercombs: Didn't feel like tracing all those threads, so I just made a new Event. There's not a specific test for the new event type, but the Nun story test does test it.

eldritch/stories.py
```diff
@@ -0,0 +1,398 @@
...
+      return self.advance_story(owner, False)
+    return None
+
+
+def CreateStories():
+  return [
+      DrifterStoryPass(), DrifterStoryFail(), DrifterStory(),
+      GangsterStoryPass(), GangsterStoryFail(), GangsterStory(),
+      NunStoryPass(), NunStoryFail(), NunStory(),
+      # Don't add stories that don't have tests
```
petercombs, line 393: I'm working on 'em, as you can see below.

eldritch/stories.py
```diff
@@ -0,0 +1,402 @@
...
+      return False
+    return None
+
+
+class PhotographerStory(Story):
+  def __init__(self):
+    super().__init__("A Thousand Words", "There's Your Proof", "The Film Is Ruined")
+
+  def get_trigger(self, event, owner, state):
+    if len([t for t in owner.trophies if isinstance(t, gates.Gate)]) >= 2:
```
fingoldfish, line 246: I think we also wanted to check the event type here, because it otherwise triggers off anything between the taking of the trophy and the completion of the story (which admittedly might be nothing).

eldritch/stories.py
```diff
@@ -0,0 +1,402 @@
...
+      return self.advance_story(owner, False)
+    return super().get_trigger(event, owner, state)
+
+
+class PhotographerStoryPass(StoryResult):
+  def __init__(self):
+    super().__init__("There's Your Proof")
+
+  def get_in_play_event(self, owner: characters.Character):
+    return events.DrawSpecific(owner, "specials", "Retainer")
```
fingoldfish, line 211: Have we set up retainers so that you can't get one if you already have one?

eldritch/stories.py
```diff
@@ -0,0 +1,402 @@
...
+
+  def get_in_play_event(self, owner: characters.Character):
+    return events.Sequence([
+        events.DiscardNamed(owner, "Retainer"),
+        events.DrawSpecific(owner, "specials", "Bad Credit")
+    ], owner
+    )
+
+  def get_override(self, other, attribute):
+    if isinstance(other, (assets.BankLoan, assets.Retainer)) and attribute == "can_keep":
```
fingoldfish, line 236: What is the difference between cancelling a KeepDrawn and changing the can_keep override? Not suggesting that you change anything, just wondering if they're functionally equivalent.

eldritch/stories.py
```diff
@@ -0,0 +1,402 @@
...
+
+  def get_usable_interrupt(self, event, owner: characters.Character, state):
+    if (
+        isinstance(event, events.Upkeep)
+        and event.character == owner
+        and owner.place.name == "Asylum"
+    ):
+      choice = events.SpendChoice(
+          owner,
+          "Spend $15 to finish story?",
```
fingoldfish, line 286: to bribe the Arkham asylum guy, presumably?

eldritch/stories.py
```diff
@@ -0,0 +1,402 @@
...
+      return self.advance_story(owner, False)
+    return None
+
+  def get_usable_interrupt(self, event, owner: characters.Character, state):
+    if (
+        isinstance(event, events.Upkeep)
+        and event.character == owner
+        and owner.place.name == "Asylum"
+    ):
+      choice = events.SpendChoice(
```
fingoldfish, line 284: Might make this a CardSpendChoice (where one of the choices is the name of the card) so that the story shows up in the user interface.

eldritch/stories.py
```diff
@@ -0,0 +1,402 @@
...
+          owner,
+          "Spend $15 to finish story?",
+          ["No", "Yes"],
+          spends=[None, values.ExactSpendPrerequisite({"dollars": 15})]
+      )
+
+      result = events.Conditional(owner, choice, "choice_idx", {1: self.advance_story(owner, True)})
```
fingoldfish, line 291: Does this need a 0: events.Nothing()?

eldritch/stories.py
```diff
@@ -0,0 +1,402 @@
...
+class SalesmanStoryPass(StoryResult):
+  def __init__(self):
+    super().__init__("Jackpot")
+
+  def get_in_play_event(self, owner: characters.Character):
+    return events.Gain(owner, {"dollars": 15}, self)
+
+  def get_usable_interrupt(self, event, owner, state):
+    if (
+        isinstance(event, events.Upkeep)
```
fingoldfish, line 305: I think you want this to interrupt SliderInput for a consistent UX

eldritch/stories.py
```diff
@@ -0,0 +1,402 @@
...
+        and event.character == owner
+        and not self.exhausted
+    ):
+      choice = events.SpendChoice(
+          owner, "Spend $3 for 1 clue token?",
+          ["No", "Yes"],
+          spends=[None, values.ExactSpendPrerequisite({"dollars": 3})]
+      )
+      cond = events.Conditional(owner, choice, "choice_idx", {1: events.Gain(owner, {"clues": 1})})
```
fingoldfish, line 314: Also need to exhaust this item.

eldritch/stories.py
```diff
@@ -0,0 +1,402 @@
...
+  def __init__(self):
+    super().__init__("It's Working!")
+
+  def get_modifier(self, other, attribute):
+    if attribute == "gate_limit":
+      return 1
+    return 0
+
+  def get_in_play_event(self, owner: characters.Character):
+    return events.AddGlobalEffect(self)
```
fingoldfish, line 361: Does this need to be a global effect? I thought cards in users' possessions would automatically be taken into account for global modifiers?

eldritch/events.py
```diff
@@ -2734,6 +2742,56 @@ def log(self, state):
     return f"[{self.character.name}] rerolled {len(self.reroll_indexes)} dice on their {ctype}"
 
 
+class RerollSpecificDice(Event):
```
fingoldfish, line 2745: Looking more closely, I think RerollSpecific will work for DiceRoll events too, if the DiceRoll is updated to have a count_successes() method.

fingoldfish (changes requested):

eldritch/stories.py
```diff
@@ -0,0 +1,202 @@
...
+    super().__init__("For a Friend", "This One's for Louis", "Crime Doesn't Pay")
+
+  def get_trigger(self, event, owner: characters.Character, state):
+    n_trophies = len([t for t in owner.trophies if isinstance(t, monsters.Monster)])
+    if isinstance(event, events.ForceTakeTrophy) and n_trophies >= 5:
+      return self.advance_story(owner, True)
+    if (
+        isinstance(event, events.GainOrLoss)
+        and (event.character == owner)
+        and (event.gains.get("dollars", 0) >= 5)
```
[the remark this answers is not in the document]
petercombs: I read it as "If he gains 5 dollars at once".  Otherwise, it would've said "If he *has* 5 dollars..." right?

eldritch/stories.py
petercombs: I forget what they were at the time, but they are now "SelfDiscardingCard"s, which means that drawing one while you already have it should allow you to skip rolling on the turn after you get it.

eldritch/stories.py
petercombs: For these cards, I think they are functionally equivalent.  I could imagine a scenario for some other character where they can no longer keep Weapons, but could keep other common/unique items, so if you were drawing multiple you wouldn't want to cancel the whole event. But I don't think there's actually any stories like that, so let's try to keep new mechanics to a minimum.

eldritch/stories.py
petercombs: > The bribe worked
so... yeah.

eldritch/stories.py
petercombs:
```python
  def globals(self) -> List[mythos.GlobalEffect]:
    return [self.rumor, self.environment, self.ancient_one] + self.other_globals
```
Looks to me like characters possessions aren't usually checked.

petercombs:
Seems possible I missed some comments, but take a look and let me know what you think!

eldritch/stories.py
petercombs: Ooh, it probably does. I see there's a TODO on the trading code to make it an event, so we can either do that as part of this PR, or leave it in the to do list.

eldritch/stories.py
fingoldfish: That's a TODO for me

eldritch/stories.py
```diff
@@ -0,0 +1,423 @@
...
+      return False
+    return None
+
+
+class ScientistStory(Story):
+  def __init__(self):
+    super().__init__("Resonance", "It's Working!", "This Is No Good")
+
+  def get_trigger(self, event, owner, state):
+    if len([t for t in owner.trophies if isinstance(t, gates.Gate)]) >= 2:
```
fingoldfish, line 406: Same as above - I think we should attach the trigger to specific event(s).

fingoldfish (changes requested):

PR #208, opened 2023-06-10
Allow ancient ones to tweak the world upon choosing them

petercombs:
This allows Squidface to reduce the maximum sanity/stamina of characters from the very beginning of the game, rather than letting them effectively have a bonus san/stam that they can't refill.

### DIFF
--- a/eldritch/eldritch.py
+++ b/eldritch/eldritch.py
@@ -772,6 +772,7 @@ class GameState:
     if ancient not in self.all_ancients:
       raise InvalidMove(f"Unknown ancient one {ancient}")
     self.ancient_one = self.all_ancients[ancient]
+    self.ancient_one.setup(self)
 
   def handle_start(self):
     if self.game_stage != "setup":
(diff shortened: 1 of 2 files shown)


petercombs:
I wasn't sure how to write a test, but running the server this has the expected behavior: choosing a character, then Squidface reduces their stats, then changing the choice to another AO restores it to the max.

fingoldfish:
I think this was already fixed in https://github.com/gummyboars/lunboks/commit/43a02beaae2663dd1d03879bec0e25a7a35073b7
Though if we need setup for other ancient ones, we can add it.

petercombs:
I think some mechanism like this will also be useful for adding Masks.

eldritch/eldritch.py
```diff
@@ -772,6 +772,7 @@ def handle_ancient(self, ancient):
     if ancient not in self.all_ancients:
       raise InvalidMove(f"Unknown ancient one {ancient}")
     self.ancient_one = self.all_ancients[ancient]
+    self.ancient_one.setup(self)
```
fingoldfish, line 775: Can I ask you to move this from handle_ancient to handle_start? If we want to do something like adding masks, we don't want to have to remove the masks when the player changes their minds about which ancient one to fight.

fingoldfish (changes requested):

fingoldfish (approved):

PR #209, opened 2023-06-11
Add the warding statue

petercombs:


eldritch/test_unique_items.py
```diff
@@ -742,6 +743,38 @@ def testKeyPreservesMovement(self):
     self.assertEqual(self.key.tokens["stamina"], 1)
 
 
+class WardingStatueTest(EventTest):
```
fingoldfish, line 746: Can you add a test where the character has a warding statue but declines to use it after failing a combat check?

fingoldfish (approved):

PR #221, opened 2023-08-22
Spy and Secretary

petercombs:


eldritch/names.js
```diff
@@ -31,6 +31,7 @@ characterData = {
   "Doctor": {"sliders": [0, 5, 0, 4, 2, 4], "focus": 2, "home": "Hospital"},
   "Archaeologist": {"sliders": [1, 3, 2, 3, 1, 5], "focus": 2, "home": "Shop"},
   "Gangster": {"sliders": [2, 4, 3, 4, 0, 3], "focus": 1, "home": "House"},
+  "Spy": {"sliders": [1,3,2,1,1,0], "focus": 1, "home": "Newspaper"},
```
petercombs, line 34: FYI, the UI is not currently representing the sliders properly... Shouldn't be an issue if the art loads, but the default drawing assumes the normal sliders and slider progression.

eldritch/characters.py
```diff
@@ -569,10 +624,11 @@ def random_possessions(self):
 
 
 def CreateCharacters():
+  from eldritch.expansions.seaside import characters as seaside_characters
```
petercombs, line 627: I assume the right way to do this is actually to allow the host to select which expansions to add in, then everything gets added in then, then character selection starts.

eldritch/expansions/seaside/abilities.py
```diff
@@ -0,0 +1,114 @@
...
+if typing.TYPE_CHECKING:
+  from eldritch.eldritch import GameState
+
+
+class AbnormalFocus(assets.Asset):
+  def __init__(self):
+    super().__init__("Abnormal Focus")
+
+  def get_interrupt(self, event, owner: characters.BaseCharacter, state):
+    if isinstance(event, events.Upkeep) and event.character == owner:
```
fingoldfish, line 17: I actually have no idea when this is supposed to happen. It could be at the start of the upkeep phase. But then the complete FAQ says that events that happen during upkeep happen during the "upkeep actions" part of upkeep, which comes after the "refresh stuff" part of upkeep. This could also mean that the Spy would have a very hard time casting spells that can only be cast during upkeep? This is very confusing.

eldritch/expansions/seaside/abilities.py
```diff
@@ -0,0 +1,114 @@
...
+          events.RemoveToken(self, "sanity", owner, self.tokens["sanity"]),
+          events.RemoveToken(self, "stamina", owner, self.tokens["stamina"]),
+      ], owner)
+    return None
+
+  def get_usable_interrupt(self, event, owner, state):
+    if (
+        not isinstance(event, events.SliderInput)
+        or event.character != owner
+        or self.get_bonus("abnormal_focus", None, owner, state) >= 3
```
fingoldfish, line 40: Because she may get an abnormal focus bonus from her personal story, I think you want to check the number of tokens on this asset instead of the total bonus.

eldritch/expansions/seaside/abilities.py
```diff
@@ -0,0 +1,114 @@
...
+        or (owner.stamina == 1 and owner.sanity == 1)
+    ):
+      return None
+    spend = events.SpendChoice(
```
fingoldfish, line 44: I was thinking of attaching the spend to the slider input for slightly fewer clicks in the UI. How are you thinking about making sure the player doesn't overspend?

eldritch/expansions/seaside/abilities.py
```diff
@@ -0,0 +1,114 @@
...
+
+class TeamPlayer(assets.Asset):
+  def __init__(self):
+    super().__init__("Team Player")
+
+  def get_usable_interrupt(self, event, owner, state):
+    in_same_place = [
+        char for char in state.characters if char != owner and char.place == owner.place
+    ]
+    if not isinstance(event, events.Upkeep) or not event.character == owner:
```
fingoldfish, line 87: Use SliderInput instead of Upkeep. The player may take actions after they've refreshed items, before they move their sliders.

eldritch/expansions/seaside/abilities.py
```diff
@@ -0,0 +1,114 @@
...
+      return None
+    choice = events.MultipleChoice(
+        owner,
+        "Grant another player +1 bonus to all skill checks until the end of the turn?",
+        ["None"] + [char.name for char in in_same_place]
+    )
+    cond = events.Conditional(
+        owner, choice, "choice_index",
+        {0: events.Nothing, **{
```
fingoldfish, line 96: Nothing() instead of Nothing

eldritch/expansions/seaside/abilities.py
```diff
@@ -0,0 +1,114 @@
...
+            i: events.DrawSpecific(char, "specials", "Team Player Bonus")
+            for i, char in enumerate(in_same_place, start=1)
+        }}
+    )
+    return events.Sequence(owner, [choice, cond])
```
fingoldfish, line 101: Don't forget to exhaust this ability? And check if it's exhausted before allowing the player to use it?

eldritch/expansions/seaside/abilities.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+class TeamPlayerBonus(assets.Card):
+  def __init__(self, idx):
+    super().__init__(
+        "Team Player Bonus", idx, "specials", {},
+        passive_bonuses={check: 1 for check in assets.CHECK_TYPES | assets.SUB_CHECKS.keys()}
```
fingoldfish, line 108: My sins have come back to haunt me. Note that a character's movement points are calculated based on only the passive speed bonuses provided by their possessions. This is because a spell like voice of ra will give +1 to speed checks, but it does NOT give +1 to movement speed, but a skill like +1 Speed does give +1 to movement speed.

So your options are to either (a) fudge this somehow (maybe this card is just always active and always provides its active bonus?), or (b) split the bonuses again: skill_bonuses, check_bonuses, active_check_bonuses.

eldritch/expansions/seaside/abilities.py
```diff
@@ -0,0 +1,114 @@
...
+    )
```
fingoldfish, line 109: Addendum: if you give +1 to fight and combat, you will actually give +2 to combat checks since they already get the fight bonus. You want just the base skill bonuses.

eldritch/values.py
```diff
@@ -382,6 +382,51 @@ def annotation(self, state):
     return f"{spend_min}-{spend_max} {self.SPEND_TYPES[self.spend_type]}"
 
 
+class FlexibleRangeSpendPrerequisite(SpendValue):
+  def __init__(self, spend_types: list, spend_min: float, spend_max: float):
+    assert all(spend_type in self.SPEND_TYPES for spend_type in spend_types)
+    super().__init__()
+    self.spend_types_ = spend_types
```
fingoldfish, line 389: Maybe `_spend_types` instead of `spend_types_`? The latter is almost never seen in python.

eldritch/values.py
```diff
@@ -382,6 +382,51 @@ def annotation(self, state):
     return f"{spend_min}-{spend_max} {self.SPEND_TYPES[self.spend_type]}"
 
 
+class FlexibleRangeSpendPrerequisite(SpendValue):
```
fingoldfish, line 385: Please cover with unit tests.

fingoldfish (changes requested):

eldritch/values.py
petercombs: Can you tell I've been doing a lot of [sklearn](https://stackoverflow.com/questions/49138692/what-does-it-mean-for-an-attribute-name-to-end-in-an-underscore) in my day job? Will fix.

eldritch/expansions/seaside/abilities.py
petercombs: What about option (c): Don't have a passive bonus, and instead get_bonus() checks the state.event_stack for a Check?

eldritch/expansions/seaside/abilities.py
petercombs: - The Spy's PassStory just lets her use the story instead of spending a sanity or stamina
- The FailStory reduces the number of Abnormal Focus available to the character (probably implemented as 
```python
def get_bonus():
    if check_type == "abnormal_focus":
        return -1
```
- `self.get_bonus` just checks the number of tokens, so works out the same, but saves me from rewriting `sum(self.tokens.values())`.

petercombs: "saves me" as if `sum(self.tokens.values())` is longer than `self.get_bonus("abnormal_focus", None, owner, state) >= 3`. 🙄

eldritch/expansions/seaside/abilities.py
petercombs: Plenty of people have argued about this, without settling on a firm conclusion. 
- https://boardgamegeek.com/thread/530872/upkeep-adjusting-skill-sliders-and-casting-spells/page/2
- https://boardgames.stackexchange.com/questions/8842/when-during-your-upkeep-can-you-cast-upkeep-spells

One resolution could be that the spy is special, and may move her sliders at the beginning of upkeep. 
> Instead, at the start of each [Upkeep Phase](http://www.arkhamhorrorwiki.com/Upkeep_Phase), move her sliders all the way to the left. She may then shift her sliders to the right a total of five shifts.

eldritch/expansions/seaside/abilities.py
petercombs: Not sure how to attach the spend to the SliderInput... I wasn't thinking about over-spending (i.e. by spending 2 at first, and then 3 a second time), but an easy work-around is to have the max_spend be `3-sum(self.tokens.values())`

eldritch/expansions/seaside/characters.py
```diff
@@ -0,0 +1,93 @@
...
+  def abilities(self):
+    return ["Abnormal Focus", "Breaking the Limits"]
+
+  def initial_attributes(self):
+    return {"dollars": 4, "clues": 2}
+
+  def fixed_possessions(self):
+    return {
+        "common": [
+            # "Cigarette Case"
```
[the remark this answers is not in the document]
petercombs: Turns out… I mostly did. Forgot to add it in CreateCommon (and I also pushed this PR before that PR was merged)

eldritch/expansions/seaside/abilities.py
fingoldfish: Upon further consideration, attaching the spend to the SliderInput seems exceptionally difficult. I think this works, even if it makes it difficult for the player to change their mind about their sliders. If it becomes a problem, we can come back and change it later.

eldritch/events.py
```diff
@@ -264,9 +264,9 @@ def resolve(self, state, name, value):  # pylint: disable=arguments-differ
       raise InvalidInput(f"Unknown slider {name}")
     if name == "done":
       if not self.free:
-        if self.character.focus_cost(self.pending) > self.character.focus_points:
+        if self.character.focus_cost(self.pending) > self.character.slider_focus_available:
           raise InvalidMove("You do not have enough focus.")
```
fingoldfish, line 268: May also need to update this error message.

eldritch/expansions/seaside/abilities.py
fingoldfish: Ah, I see. I thought it gave her one extra shift, but it needs to be exhausted to give her one extra shift. That is a small, but important distinction that I did not pick up on.

eldritch/expansions/seaside/abilities.py
```diff
@@ -0,0 +1,122 @@
...
+
+class Synergy(assets.Asset):
+  def __init__(self):
+    super().__init__("Synergy")
+
+  def get_bonus(
+      self, check_type, attributes, owner: characters.BaseCharacter, state: "GameState"
+  ):
+    # TODO: Call Stack enforcement to guard against infinite loops.
+    n_allies = len([ally for ally in owner.possessions if ally.deck == "allies"])
```
fingoldfish, line 74: You'll want to use `getattr(ally, "deck", "")` since not every asset has a `deck` attribute.

eldritch/expansions/seaside/abilities.py
```diff
@@ -0,0 +1,122 @@
...
+    n_in_same_place = len(
+        [char for char in state.characters if char != owner and char.place == owner.place]
+    )
+    if n_allies == 0 and n_in_same_place == 0:
+      return 0
+    if check_type in (assets.CHECK_TYPES | assets.SUB_CHECKS.keys()):
```
fingoldfish, line 80: Can you test this with a spell check? I think it might grant 2 bonus instead of 1.

eldritch/expansions/seaside/abilities.py
fingoldfish: I feel (c) might be more brittle than (a), but I don't have any examples I can point to. I think it works?

eldritch/expansions/seaside/abilities.py
fingoldfish: I think the new implementation may have the same issue?

eldritch/values.py
fingoldfish: I can't tell, but only because I've not heard of sklearn until now.

eldritch/expansions/seaside/characters.py
```diff
@@ -0,0 +1,87 @@
...
+    return {"dollars": 4, "clues": 2}
+
+  def fixed_possessions(self):
+    return {"unique": ["Yellow Play"]}
+
+  def random_possessions(self):
+    return {"common": 2, "unique": 1, "spells": 1, "skills": 1}
+
+
+class Spy(characters.BaseCharacter):
```
fingoldfish, line 21: Can you add some tests that cover the upkeep for the Spy (and maybe Secretary)?

fingoldfish:
Few more things, that will probably also be uncovered by unit tests. Getting close.

fingoldfish (changes requested):
I still haven't figured out how to review pull requests properly.

eldritch/characters.py
```diff
@@ -219,65 +273,66 @@ def get_spend_event(self, handle):
...
-          bonus = (bonus + 1) // 2
-      modifier += bonus
-    return modifier
+  def focus_cost(self, pending_sliders):
+    return sum(abs(orig - pending_sliders[name]) for name, orig in self.sliders().items())
 
-  def get_modifier(self, other, attribute):
-    return sum(p.get_modifier(other, attribute) or 0 for p in self.possessions)
+  @property
+  def slider_focus_available(self):
```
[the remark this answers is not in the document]
petercombs: Methodized.

eldritch/expansions/seaside/abilities.py
```diff
@@ -0,0 +1,127 @@
...
+    if check_type in assets.CHECK_TYPES:
+      return 1
+    return 0
+
+
+class TeamPlayer(assets.Asset):
+  def __init__(self):
+    super().__init__("Team Player")
+
+  def get_usable_interrupt(self, event, owner, state):
```
fingoldfish, line 94: Do we want this to be unconditional? If the player can always choose None, then maybe we always present them with a prompt? It's probably fine as-is, just thinking out loud.

eldritch/expansions/seaside/abilities.py
```diff
@@ -0,0 +1,127 @@
...
+    in_same_place = [
+        char for char in state.characters if char != owner and char.place == owner.place
+    ]
+    if not isinstance(event, events.SliderInput) or not event.character == owner or self.exhausted:
```
fingoldfish, line 98: I think you also want to make sure the SliderInput event is not done.

eldritch/expansions/seaside/abilities.py
```diff
@@ -0,0 +1,127 @@
...
+  def __init__(self, idx):
+    super().__init__("Team Player Bonus", idx, "specials", {}, {})
+
+  def get_bonus(self, check_type, attributes, owner, state):
+    if len(state.event_stack) > 0 and isinstance(state.event_stack[-1], events.Check):
+      return 1
+    return 0
+
+  def get_interrupt(self, event, owner, state):
+    if isinstance(event, events.Mythos) and event.is_done():
```
fingoldfish, line 125: If we're looking for a Mythos that's done, shouldn't this be a trigger instead of an interrupt?

eldritch/expansions/seaside/characters.py
```diff
@@ -0,0 +1,86 @@
+from eldritch import characters
+
+
+class Secretary(characters.Character):
```
fingoldfish, line 4: If you're adding the secretary here, can you also add her to names.js, and add the specials to the specials deck?

eldritch/expansions/seaside/test_abilities.py
```diff
@@ -0,0 +1,99 @@
...
+    self.assertTrue(self.state.usables)
+    self.state.event_stack.append(usable)
+    choice = self.resolve_to_choice(events.MultipleChoice)
+    self.spend("sanity", 1, choice)
+    choice.resolve(self.state, "Yes")
+    sliders = self.resolve_to_choice(events.SliderInput)
+    self.assertEqual(self.char.slider_focus_available(), 8)
+    self.assertFalse(self.state.usables)
+
+    with self.assertRaisesRegex(InvalidMove, "enough focus"):
```
fingoldfish, line 92: Nit: put this with assertraises around just the last slider movement.

eldritch/expansions/seaside/test_abilities.py
```diff
@@ -0,0 +1,99 @@
...
+    self.state.all_characters["Nun"] = nun
+    self.state.characters.append(nun)
+
+    sliders = events.SliderInput(self.char)
+    self.state.event_stack.append(sliders)
+    may_use = self.resolve_to_usable(0, "Team Player")
+    self.state.event_stack.append(may_use)
+
+    team = self.resolve_to_choice(events.MultipleChoice)
+    self.assertListEqual(team.choices, ["None", "Nun"])
```
fingoldfish, line 51: No pun intended.

eldritch/expansions/seaside/test_abilities.py
```diff
@@ -0,0 +1,99 @@
...
+from eldritch.expansions.seaside import abilities
+from eldritch.expansions.seaside import characters as seaside_characters
+from eldritch import assets
+from eldritch import characters
+from eldritch import events
+from eldritch.items import spells
+
+
+class TestSecretaryAbilities(EventTest):
+  def testSynergy(self):
```
fingoldfish, line 13: Add one test for a character in the same area (as opposed to the ally)?

fingoldfish (changes requested):
Just one or two more things.

eldritch/expansions/seaside/test_abilities.py
petercombs: Nun at all...

fingoldfish (approved):

eldritch/items/__init__.py
```diff
@@ -1,20 +1,26 @@
...
 
 def CreateTradables():
   return [DeputysRevolver(), PatrolWagon()]
 
 
 def CreateSpecials():
-  return (
-      [Deputy()]
+  ret: List[Asset] = (
+      [Deputy(), TeamPlayerBonus(0)]
```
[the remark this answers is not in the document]
petercombs: Almost! I need to put it in abilities.py.

PR #227, opened 2023-09-10
Fix some item bugs

petercombs:
- A rare interaction with the Elder Thing and Enchant Weapon can enable a glitch where you permanently make a physical weapon magical.
- Silver Key is usable after checks that you pass
- The effect of Voice of Ra should(?) remain active even after you no longer have the spell.

### DIFF
--- a/eldritch/characters.py
+++ b/eldritch/characters.py
@@ -140,11 +140,11 @@ class BaseCharacter(metaclass=abc.ABCMeta):
   def bonus(self, check_name, state, attributes=None, check=True):
     modifier = 0
     if state:
-      modifier += state.get_modifier(self, check_name)
+      modifier += state.get_modifier(self, check_name + ("_check" if check else ""))
     for pos in self.possessions:
-      bonus = pos.get_bonus(check_name, attributes)
+      bonus = pos.get_bonus(check_name, attributes, self, state)
       if check:
-        bonus += pos.get_bonus(check_name + "_check", attributes)
+        bonus += pos.get_bonus(check_name + "_check", attributes, self, state)
       if attributes and check_name in {"magical", "physical"}:
         if check_name + " immunity" in attributes:
           bonus = 0
@@ -335,10 +335,10 @@ class Character(BaseCharacter):
     self.movement_points = self._speed_sneak[self.speed_sneak_slider][0]
 
   def max_stamina(self, state):
-    return self._max_stamina + self.bonus("max_stamina", state)
+    return self._max_stamina + self.bonus("max_stamina", state, check=False)
 
   def max_sanity(self, state):
-    return self._max_sanity + self.bonus("max_sanity", state)
+    return self._max_sanity + self.bonus("max_sanity", state, check=False)
 
   def base_speed(self):
     return self._speed_sneak[self.speed_sneak_slider][0]
--- a/eldritch/characters.py
+++ b/eldritch/characters.py
@@ -65,43 +65,43 @@ class BaseCharacter(metaclass=abc.ABCMeta):
   def base_speed(self):
     raise NotImplementedError
 
-  def speed(self, state):
-    return self.base_speed() + self.bonus("speed", state)
+  def speed(self, state, check=True):
+    return self.base_speed() + self.bonus("speed", state, check=check)
 
   @abc.abstractmethod
   def base_sneak(self):
     raise NotImplementedError
 
-  def sneak(self, state):
-    return self.base_sneak() + self.bonus("sneak", state)
+  def sneak(self, state, check=True):
+    return self.base_sneak() + self.bonus("sneak", state, check=check)
 
   @abc.abstractmethod
   def base_fight(self):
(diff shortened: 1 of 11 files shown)


eldritch/characters.py
```diff
@@ -219,10 +219,12 @@ def get_spend_event(self, handle):
       return events.ReturnMonsterToCup(self, handle)
     return events.ReturnGateToStack(self, handle)
 
-  def bonus(self, check_name, state, attributes=None):
+  def bonus(self, check_name, state, attributes=None, check=True):
     modifier = state.get_modifier(self, check_name)
     for pos in self.possessions:
       bonus = pos.get_bonus(check_name, attributes)
+      if check:
```
fingoldfish, line 226: I think you will need to also do the following:
- update the lantern to only give +1 to luck checks
- update all mythos cards that affect checks
- update the Check event to use the new _check attribute

I'm also open to other ideas on how to distinguish between checks/raw attributes.

fingoldfish (changes requested):

eldritch/events.py
```diff
@@ -2328,22 +2331,31 @@ def resolve(self, state):
...
+    if self.discarded is None:
+      self.discarded = []
+      for item in self.items:
+        for attr in item.tokens:
+          item.tokens[attr] = 0
+        if item not in self.character.possessions:
+          continue
+        self.character.possessions.remove(item)
+        item._exhausted = False  # pylint: disable=protected-access
+        if hasattr(item, "_active"):
```
fingoldfish, line 2343: Since we are now explicitly deactivating the items, I think this if block can be removed?

eldritch/events.py
```diff
@@ -2417,24 +2429,28 @@ def __init__(self, character, item_name):
...
-        deck.append(item)
-        self.discarded = item
-        return
-    self.discarded = False
+    if self.discarded is None:
+      for item in self.character.possessions:
+        if item.name == self.item_name:
+          self.character.possessions.remove(item)
+          item._exhausted = False  # pylint: disable=protected-access
+          if hasattr(item, "_active"):
```
fingoldfish, line 2440: Same comment here.

eldritch/items/spells.py
```diff
@@ -170,6 +170,20 @@ def get_usable_interrupt(self, event, owner, state):
     )
     return events.CastSpell(owner, self, choice=choice)
 
+  def get_interrupt(self, event, owner, state):
+    if (
+        isinstance(event, (events.DiscardSpecific, events.DiscardNamed))
+        and event.character == owner
+        and event.discarded
+        and self in event.discarded
```
fingoldfish, line 178: For DiscardNamed, the `event.discarded` will be the individual item, so this `in` check will fail. This is probably on me, as I never should have made those two different events, but we should probably work around it for now.

eldritch/items/spells.py
```diff
@@ -170,6 +170,20 @@ def get_usable_interrupt(self, event, owner, state):
...
+    ):
+      # TODO: Should we figure out a way to have the effect last until the end of combat?
+      # The only way I can think of to lose an item during combat is the Elder Thing, which is not
+      # resistant. Plus, the rules are non-specific as to whether discarding a spell makes its
+      # effect go away.
+      return events.DeactivateSpell(owner, self)
+    return super().get_trigger(event, owner, state)
```
fingoldfish, line 185: I think you meant to call `super().get_interrupt` instead of `get_trigger`

eldritch/test_combat.py
```diff
@@ -2722,6 +2722,62 @@ def testCannotUseEnchantWeaponInNextCombat(self):
       self.resolve_until_done()
       self.assertEqual(rand.call_count, 2)  # Fight (4) + spawn (-2) + physical immunity (0)
 
+  def testElderThingGlitch(self):
+    elder_thing = monsters.ElderThing()
+    self.state.event_stack.clear()
```
fingoldfish, line 2727: You have a duplicate event stack clear here

eldritch/test_combat.py
```diff
@@ -2722,6 +2722,62 @@ def testCannotUseEnchantWeaponInNextCombat(self):
...
+      fight_or_evade = self.resolve_to_choice(FightOrEvadeChoice)
+      fight_or_evade.resolve(self.state, "Fight")
+      choose_weapons = self.resolve_to_choice(CombatChoice)
+    self.assertTrue(self.char.possessions[2].active)
+    self.assertEqual(self.char.hands_available(), 2)  # Enchant weapon is handless.
+
+    # We already cast on the revolver, now choose to use it
+    self.choose_items(choose_weapons, [".38 Revolver0"])
+
+    with mock.patch.object(events.random, "randint", new=mock.MagicMock(return_value=1)) as rand:
```
fingoldfish, line 2753: Add a comment that we are losing combat with the elder thing

fingoldfish (approved):
LGTM after fixing type error for `self.discarded`

PR #236, opened 2024-03-17
Add Shaman character

shruti-temp:


eldritch/characters.py
```diff
@@ -647,13 +647,31 @@ def random_possessions(self):
...
+      return [] # TODO: Guardian of the Veil
+
+  def initial_attributes(self):
+    return {"dollars": 3, "clues": 1}
+
+  def fixed_possessions(self):
+    return {}
+
+  def random_possessions(self):
+    return {"unique": 1, "spells": 2, "n_gate_trophies": 1, "skills": 1}
```
fingoldfish, line 665: Unfortunately, n_gate_trophies is a readonly property of the character, and character initialization code requires that the keys here be names of decks. I think you will need to update the character initialization code to understand what this is and how to give a gate trophy at the start of the game.

fingoldfish (changes requested):

PR #240, opened 2024-07-14
Farmhand and Urchin code and tests

petercombs:


eldritch/expansions/clifftown/test_clifftown_abilities.py
```diff
@@ -0,0 +1,88 @@
...
+    self.assertFalse(arrested.cancelled)
+
+  def testMinorCredit(self):
+    self.char.place = self.state.places["Bank"]
+    self.advance_turn(0, "encounter")
+    self.resolve_until_done()
+
+    self.char.possessions.remove(self.minor)
+    self.advance_turn(1, "encounter")
+    self.resolve_until_done()
```
petercombs, line 88: This test isn't working as I anticipated... do the FixedEncounters not get loaded when in test mode?

eldritch/expansions/seaside/abilities.py
```diff
@@ -144,3 +144,40 @@ def get_trigger(self, event, owner, state):
...
     return None
+
+
+class ThickSkulledCombat(events.Combat):
+  def resolve(self, state):
+      # Horror check
+    if (
+        self.monster.difficulty("horror", state, self.character) is not None
+        and self.horror is None
+        and (self.evade is not None or self.combat is not None)
```
petercombs, line 155: This added condition is new compared to regular combat.

eldritch/events.py
```diff
@@ -3874,30 +3874,42 @@ def __init__(self, character, monster):
...
-        if self.horror.success and nightmarish:
-          self.sanity_loss = Loss(
-              self.character, {"sanity": self.monster.bypass_damage("horror", state)})
-          state.event_stack.append(self.sanity_loss)
-          return
 
     # Combat or flee choice.
+    self._do_combat_or_evade(state)
+
+  def _setup_horror(self, state):
```
petercombs, line 3886: Spinning off to separate methods makes it easier to do ThickSkulledCombat.

eldritch/expansions/seaside/abilities.py
```diff
@@ -144,3 +144,40 @@ def get_trigger(self, event, owner, state):
...
+    super().__init__("Thick Skulled")
+
+  def get_interrupt(self, event, owner, state):
+    if (
+        isinstance(event, events.Combat)
+        and not isinstance(event, ThickSkulledCombat)
+        and event.character == owner
+    ):
+      print("matched")
+      return events.Sequence([
```
fingoldfish, line 179: So here's the list of things that I think we would want to test since we're cancelling the original combat:
- combat with dimensional shambler and/or nightgaunt, which check the event stack and try to cancel the evadeorcombat. might need to have a second monster in the same location to make sure the evadeorcombat is handled properly.
Actually, I think that's it for now. There will likely be more as we add masks, e.g.
http://www.arkhamhorrorwiki.com/The_Bloated_Woman

eldritch/expansions/clifftown/test_clifftown_abilities.py
fingoldfish: Can you define not working as anticipated? See other comments about can_take vs can_get for my theory on why this may not be working as expected.

eldritch/test_mythos.py
```diff
@@ -1040,6 +1040,44 @@ def testSealChoiceCancelled(self):
...
+    self.state.event_stack.append(close)
+
+    elder_sign = self.resolve_to_usable(0, "Elder Sign0")
+    self.state.event_stack.append(elder_sign)
+    self.resolve_until_done()
+
+    self.assertIsNone(self.square.gate)
+    self.assertTrue(self.square.sealed)
+    self.assertEqual(self.char.sanity, starting_sanity - 1)
+    self.assertEqual(self.char.stamina, starting_stamina - 1)
```
fingoldfish, line 1057: Also start with a doom token and assert it gets removed?

eldritch/test_mythos.py
```diff
@@ -1040,6 +1040,44 @@ def testSealChoiceCancelled(self):
...
+    close = GateCloseAttempt(self.char, "Square")
+    self.state.event_stack.append(close)
+
+    elder_sign = self.resolve_to_usable(0, "Elder Sign0")
+    self.state.event_stack.append(elder_sign)
+    self.resolve_to_choice(events.ItemLossChoice)
+
+    self.assertIsNone(self.square.gate)
+    self.assertTrue(self.square.sealed)
+    self.assertEqual(self.char.sanity, 1)
```
fingoldfish, line 1079: Also assert that the gate still ends up in the character's trophies?

fingoldfish (changes requested):
Mostly LGTM, a few small things before it's good to go.

eldritch/expansions/clifftown/characters.py
```diff
@@ -0,0 +1,24 @@
+from eldritch import characters
+
+
+class Urchin(characters.Character):
```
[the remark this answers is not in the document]
petercombs: Fun fact, [urchin](https://en.wikipedia.org/wiki/Urchin) originally meant hedgehog, so a sea urchin is a sea-hedgehog.

eldritch/values.py
```diff
@@ -156,6 +157,20 @@ def value(self, state):
     return sum(item.name == self.item_name for item in self.character.possessions) == 0
 
 
+class OverridePrerequisite(Value):
+  def __init__(self, character: characters.Character, attribute: str, other=None):
+    super().__init__(error_fmt=f"State does not allow '{attribute}'")
```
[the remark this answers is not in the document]
petercombs: How about this approach of just letting the creating object specify what the message should be (but autocalculating if not specified).

eldritch/expansions/clifftown/abilities.py
```diff
@@ -0,0 +1,52 @@
...
+    # Minor, Not Miner
+    super().__init__("Minor")
+
+  def get_interrupt(self, event, owner, state):
+    if isinstance(event, events.TakeBankLoan) and event.character == owner:
+      return events.CancelEvent(event)
+    return None
+
+  def get_override(self, other, attribute):
+    if attribute == "can_get_bank_loan":
```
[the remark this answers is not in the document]
petercombs: I unified everything to "can_get_bank_loan".  It seemed possible that there's an environment that forbids loans, but I can't find one. Up to you whether to go back to the old "do you have bad credit" check.

eldritch/expansions/seaside/abilities.py
petercombs: ~Nightgaunt~ Dream Flier test aded.

eldritch/expansions/clifftown/test_clifftown_abilities.py
petercombs: When the DummyCharacter is no longer a minor (happy birthday?), I was anticipating them to be offered a choice to either take a loan or have a regular encounter. But the fact that they can resolve_until_done suggests I'm misunderstanding something.

eldritch/expansions/seaside/abilities.py
```diff
@@ -144,3 +144,40 @@ def get_trigger(self, event, owner, state):
...
+
+
+class ThickSkulledCombat(events.Combat):
+  def resolve(self, state):
+      # Horror check
+    if (
+        self.monster.difficulty("horror", state, self.character) is not None
+        and self.horror is None
+        and (self.evade is not None or self.combat is not None)
+        and any([self.evade.evaded is False, self.combat.defeated is False])
```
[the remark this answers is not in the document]
petercombs: Normally, self.evade and self.combat get filled in at the same time, so it doesn't much matter whether we do `and` or `or`. That said, getattr is definitely the safest.

eldritch/expansions/seaside/test_abilities.py
```diff
@@ -6,9 +6,104 @@
...
+    self.assertIsNone(combat.horror)
+    self.assertEqual(self.char.sanity, 5)
+
+    fight_or_flee.resolve(self.state, "Fight")
+    choose_weapons = self.resolve_to_choice(events.CombatChoice)
+    self.choose_items(choose_weapons, [])
+
+    # Fail the combat check, go through the gate
+    with mock.patch.object(events.random, "randint", new=mock.MagicMock(return_value=3)):
+      # next_fight_or_flee = self.resolve_to_choice(events.FightOrEvadeChoice)
```
fingoldfish, line 99: Can remove comment

eldritch/values.py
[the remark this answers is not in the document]
fingoldfish: LGTM

eldritch/expansions/clifftown/test_clifftown_abilities.py
fingoldfish: Oh, you're correct - fixed encounters do not get initialized in test mode. You'll need to add these three lines https://github.com/gummyboars/lunboks/blob/master/eldritch/test_game.py#L55-L57
to your setUp() to make it work.

fingoldfish (approved):

